PR #77, opened 2022-05-30
Issue68 change app name inside toolbar

### LINKED ISSUE #68: Change app name inside browser window tab
Now the name inside the app window tab is 'Svelte app'. It should be changed to the web app name ('Ocean farm' maybe)

KatunaNorbert:
Fixes #68.

- [x] change app name inside toolbar
- [x] dynamically change toolbar title to reflect current page
- [x] redirect invalide pathname to claim rewards page

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4,8 +4,6 @@
 	<meta charset='utf-8'>
 	<meta name='viewport' content='width=device-width,initial-scale=1'>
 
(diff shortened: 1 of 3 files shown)


idiom-bytes (approved):
Looks great. Not sure if the name should be "Ocean Farm" or "Data Farming".
Let's get PR pushed, and update as needed.

PR #78, opened 2022-05-30
Feature/39 stake in pools

### LINKED ISSUE #39: Pool explorer: can stake in pools
What: each row in the pools explorer has a “stake/unstake” button. When pressed, user has a means to add stake or remove stake.

idiom-bytes:
Fixes #39 

Changes proposed in this PR:
- Implemented modal
- Implementing switch network functionality inside of modal
- Hooked up basic token, bpool contracts, in addition to basic views & calls
- Got staking to work, however am now running into an error while trying to retrieve the 1SS staked event results.

TODO:
DONE - Fix any issues, or implement any additional functionality to complete staking. (i.e. % of pool ~= fn as market).
DONE - Amount staked is currently not being returned because receipt events are empty.
DONE - Merge w/ sql queries & current pools from v4 marketplace
- Update rewards distribution in vm


idiom-bytes:
I would like to propose we submit this PR and then fix a few other things. I'll create tickets for these.

[Redundant Approval] - Approval how it's implemented right now can be redundant. It should check if it's already approved, and update the amounts/estimates, and try to complete the staking based on existing amount approved for. No option for forcing override atm.

[Close Button] - Disable close button from StakeModal if "loading" state is in place. Right now, loading has been abstracted out of the Staking component, so they are detached and need to be wired together.

[Ropsten] - Ropsten seems to be down due to people messing with it. Maybe we should use a different testnet, or a different chain's testnet (i.e. polygon testnet)

[Pools Table & Columns] - There are various columns that we should not show inside of the Pools table. Create a default filter for those columns. Further, right now the LP dictionary inside of pools, seems redundant. This could all be cleaned up a bit better I believe. Also, table is currently in commons, but it's not a generic table (i.e. it has a lot of pool-related functionality).

[SQL & Data from Subgraph] - Grab the full DT name, if we could grab the metadata (data title), it would be more user-friendly. Right now we have a bunch of low-level data that isn't as useful.

[Pool Explorer - TVL & Total Assets] - Total Assets makes sense because they are fungible. TVL is not because the base data token can be in different chains, and can be different base token altogether. If we had these values in USD, then that would be more applicable. Surface this data to the top of the pools explorer table.

KatunaNorbert (approved):

KatunaNorbert:
Good job @idiom-bytes !
I just did some clean up by removing some commented code, unused import and console logs.
I agree with this issue being a complex one and that we should merge it as it is functional and handle other improvements in new issues, so I'm going to merge this.

PR #89, opened 2022-06-07
Issue87 split stake and approve

### LINKED ISSUE #87: Improve Staking - addDTLiquidity => Approve + Stake
This ticket is here so we can push the PR (it's getting huge) and get them clean.

Problem:
We implemented addDTLiquidity. This works however it can require the user to approve transaction twice (it should only request approval once). Further, we might want to implement like the marketplace where the user must first approve, and then stake (not addDTLiquidity).

Future Improvement:
- Add "approve button" if amount > allowance
- Break down staking into Approve + Stake
- Enable user to force override approval w/ checkbox

### DoD:
- Approval checks if allowance is greater, and uses that instead
- User is not required to approve twice if balance selected < previously approved

KatunaNorbert:
Fixes #87.

### DIFF
--- a/src/components/common/TokenApproval.svelte
+++ b/src/components/common/TokenApproval.svelte
@@ -1,0 +1,47 @@
+<script>
+  import Button from "./Button.svelte";
+  import { approve as approveToken } from "../../utils/tokens";
+  import { networkSigner, userAddress } from "../../stores/web3";
+  import { isTokenAmountApproved } from "../../utils/tokens";
+
+  export let disabled = false;
+  export let loading = false;
+  export let amount;
+  export let tokenName;
+  export let tokenAddress;
+  export let poolAddress;
+
+  let isAmountApproved;
+
+  const onClick = async () => {
+    loading = true;
+    approveToken(tokenAddress, poolAddress, spender, amount, $networkSigner);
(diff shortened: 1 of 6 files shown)


src/components/common/TokenApproval.svelte
```diff
@@ -0,0 +1,73 @@
...
+      }
+    );
+  };
+
+  $: amount,
+    isTokenAmountApproved(
+      tokenAddress,
+      amount,
+      $userAddress,
+      poolAddress,
```
idiom-bytes, line 51: Great! Awesome to see you added this and expanded the UX around it!

src/utils/tokens.js
```diff
@@ -66,44 +64,30 @@ export const allowance = async (
...
   }
-
   try {
       // TODO - Override gas price & limit
       // let gasPrice = getFairGasPrice();
-      const tx = await datatoken.approve(
-          spender,
-          ethers.utils.parseEther(amount.toString()));
+      const tx = await datatoken.approve(spender,ethers.utils.parseEther(amount.toString()));
+      await tx.wait()
```
idiom-bytes, line 87: Nice catch. Thank you!

PR #142, opened 2022-07-01
Better explain TVL and Volume

### LINKED ISSUE #105: Pool explorer: help user know what "TVL" and "Volume" columns mean
**Motivation**

Questions that users may ask:
- What units are TVL and volume in? Eg OCEAN, H2O, USD, other?
- Does TVL include the value of datatoken, or basetoken+datatoken?
- What is "volume"?
- Over what time interval is volume? Eg the last day, the last week, since counting started for the current round, or lifetime for the asset?

Right now Pool Explorer doesn't answer these.

**Suggested solution**

- Put "$" text to the left "TVL" text
- Put "$" text to the left "Volume" text
- Have a tooltip (tiny "?" next to the column label) to the right of "TVL" text. When mouse hovers above, it says "Total value locked in basetoken (OCEAN, etc) for this pool, in units of USD"
- Have a tooltip to the right of "Volume" text. When mouse hovers above, it says "Data consume volume for this pool, in units of USD, since the start of this DF round". I'm assuming that ", since the start of this DF round" is the interval you're showing, if it's something else, then revise text accordingly (or don't show that part)

KatunaNorbert:
Fixes #105.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ SUPPORTED_CHAIN_IDS = [3, 4, 1287, 80001]
 POOLS_API = https://test-df-sql.oceandao.org/pools
 
 //rewards config containing the airdrop contracts and token addresses for each network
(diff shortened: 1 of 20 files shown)


idiom-bytes:
Looks like the commits have already made into main, but the stack/logs are cleared.
I'm not sure if there is anything else to be done here.

KatunaNorbert:
The part with adding the currency beside the value was handled, but I was stock with adding the tooltip as far as I remember. There was some issue with table not allowing to send components as cell values I think, let's keep it open and will take another look.

KatunaNorbert:
@idiom-bytes  This one is deprecated. We already have symbols next to values and then we have another issue for adding tooltips where we can also handle the extra explanation of table columns

PR #196, opened 2022-08-26
Feature/195 update ve ocean lock cards

### LINKED ISSUE #195: FE - Update live calculations of OCEAN lockup, multiplier, and voting power.
### DoD:
- [x] Changing dates updates estimated values
- [x] Switch apy into estimated voting power
- [x] Get multiplier to properly update when dates change
- [x] Update locking timelines (made them closer to what CRV offers in their FE)
- [x] Made max 4 yrs

idiom-bytes:
Fixes #195
- ordered items to better communicate time bonus vs. voting power
- Improved estimated veOcean
- made titles and other descriptions easier to reduce complexity
- removed APY calculation

### DIFF
--- a/src/components/common/DateInput.svelte
+++ b/src/components/common/DateInput.svelte
@@ -1,4 +1,6 @@
 <script>
+  import { getThursdayDate } from "../../utils/functions";
+
   export let value = undefined;
   export let placeholder = undefined;
   export let min = undefined;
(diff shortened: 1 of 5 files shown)


KatunaNorbert (approved):
I did some fixes so the time bonus is 100% and Voting Power equals to amount to lock when the selected period is 4 years.
Also included the fixes for #190 in here, which made thinks easier to validate.

PR #203, opened 2022-08-31
Update claim rewards and other components to be single network

### LINKED ISSUE #202: Refactor claim rewards page
Claim rewards page should be update to only handle the rewards for 1 chain instead of multiple chains

KatunaNorbert:
Fixes #202.

### DIFF
--- a/src/components/veocean/WithdrawOcean.svelte
+++ b/src/components/veocean/WithdrawOcean.svelte
@@ -47,10 +47,17 @@
     />
   </div>
   <div class="item">
-    <p>
-      <strong>OCEAN</strong> withdrawal will be enabled as soon as the lock period
-      is over.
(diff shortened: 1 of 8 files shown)


src/components/veocean/WithdrawOcean.svelte
```diff
@@ -47,10 +47,17 @@
...
+      <p>
+        <strong>OCEAN</strong> withdrawal will be enabled as soon as the lock period
+        is over.
+      </p>
+    {:else}
+      <p>
+        No <strong>OCEAN</strong> locked. Lock your <strong>OCEAN</strong> tokens
+        first.
+      </p>
+    {/if}
```
idiom-bytes, line 60: Good catch.

Issue #213, opened 2022-09-08
When error messages are thrown, mostly due to contract/configuration, our infura key is getting logged


idiom-bytes:
I noticed in the staging environment, that if the contract address is wrong our infura key shows up.

How to test:
- Change the contract address to something wrong, you'll see the log in the console

PR #223, opened 2022-09-13
Updated ABIs, fixed delegations, implemented fee estimates and update…

idiom-bytes:
Updates UX/UI to be similar to figma, connects to contracts inside rinkeby.
- The reward nomenclature is not 1:1 across the different reward systems
- DFRewards provides an Estimated Reward + Claimable Reward
- VERewards only provides an estimateClaim reward

Screenshot from 2022-09-12 22-35-34

### DIFF
--- a/src/components/common/ClaimItem.svelte
+++ b/src/components/common/ClaimItem.svelte
@@ -1,62 +1,49 @@
 <script>
+  import Button from "./Button.svelte";
+  import ItemWithLabel from "./ItemWithLabel.svelte";
+
   export let title;
   export let amount;
   export let loading = false;
   export let onClick;
 
-  const onClaimClicked = () => {
+  const onClaimClicked = async () => {
     loading = true;
-    onClick()
-  }
+    await onClick();
+    loading = false;
+  };
 </script>
 
-{#if loading === false}
-  <div class="container" on:click={() => onClaimClicked()}>
-    <div class="description">
-      <span class="title">{title}</span>
-      <span class="value">${amount}</span>
-    </div>
-    <div class="description">
-      <span class="claim">Claim</span>
-    </div>
+<div class="container">
+  <ItemWithLabel {title} value={amount} />
+  <div class="actionContainer">
+    <Button
+      text={loading ? "Loading..." : "Claim"}
+      onClick={() => onClaimClicked()}
+      textOnly
+      disabled={loading}
+    />
   </div>
-{:else}
-  <div class="container">
-    <div class="description">
-      <span class="value">Loading</span>
-    </div>
-  </div>
-{/if}
-
+</div>
 
 <style>
   .container {
     display: flex;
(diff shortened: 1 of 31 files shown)


src/stores/claimables.js
```diff
@@ -0,0 +1,6 @@
+import { writable } from "svelte/store";
+
+export let veEstimate = writable(undefined);
+export let veClaimables = writable(undefined);
+export let dfEstimate = writable(undefined);
+export let dfClaimables = writable(undefined);
```
KatunaNorbert, line 6: I think we can merge the content from this file with `airdrops.js` and have just one store for all rewards.

src/utils/feeEstimate.js
```diff
@@ -0,0 +1,18 @@
...
+    try {
+      const contract = new ethers.Contract(process.env.VE_FEE_ESTIMATE_CONTRACT, veFeeEstimateABI, get(networkSigner));
+      const estimateClaim = await contract.estimateClaim(userAddress)
+      const estimateClaimFormatted = ethers.utils.formatEther(BigInt(estimateClaim).toString(10))
+      return estimateClaimFormatted
+  } catch (error) {
+    console.log(error)
+    throw error;
+  }
+}
```
KatunaNorbert, line 18: We could also move this content over to a common rewards file since it would be just one function in this file

src/components/common/ClaimItem.svelte
```diff
@@ -0,0 +1,62 @@
...
+{#if loading === false}
+  <div class="container" on:click={() => onClaimClicked()}>
+    <div class="description">
+      <span class="title">{title}</span>
+      <span class="value">${amount}</span>
+    </div>
+    <div class="description">
+      <span class="claim">Claim</span>
+    </div>
+  </div>
```
KatunaNorbert, line 22: We could use our existing components to create this one. On the left side we can use the `ItemWithLabel` component and for the claim part user the `Button` component

src/components/common/ClaimItem.svelte
```diff
@@ -0,0 +1,62 @@
...
+{:else}
+  <div class="container">
+    <div class="description">
+      <span class="value">Loading</span>
+    </div>
+  </div>
```
KatunaNorbert, line 28: I will suggest this loading element will only replace the button instead of the entire content

KatunaNorbert (changes requested):

KatunaNorbert:
I merged the main into this branch bringing the changes from #213 and also updated the new contract calls to make it compatible with those changes

src/utils/feeEstimate.js
idiom-bytes: I'm going to suggest we leave this like this, because the `src/utils/` file-pattern is to maps fns 1:1 w/ the ABI

PR #225, opened 2022-09-13
Issue214 get allocations from subgraph

### LINKED ISSUE #214: User Allocations - Live vs. Calculated metrics via subgraph & dfsql
DFSQL updates every 5 minutes.
If the user allocates and we are consuming from dfsql, it will not show the user latest allocation.
We should retrieve any "live" metrics directly from the subgraph, while querying "colder" metrics from dfsql

Table Logic:
1. Query allocations from subgraph + cache it
1. Query datanfts from dfsql
1. Join allocation and datanft data
1. Feed to table + sort by allocation

Generally:
- To get data, retrieve from dfsql (updates every 5 mins)
- To get specific values, retrieve from subgraph (nearly live)

### DoD:
- [ ] When user allocates more, cache it
- [ ] When the application reloads, or the page reloads, fetch allocations again from subgraph
- [ ] Table is serving data from the right sources and is updated nearly-live rather than every 5 minutes

KatunaNorbert:
Fixes #214.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,10 +9,12 @@
       "version": "1.0.0",
       "dependencies": {
         "@rollup/plugin-replace": "^4.0.0",
+        "apollo-boost": "^0.4.9",
         "carbon-components-svelte": "^0.63.8",
         "carbon-icons-svelte": "^11.0.1",
         "ethers": "^5.6.5",
         "sirv-cli": "^2.0.0",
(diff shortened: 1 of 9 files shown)


idiom-bytes:
"Good to go"

Ways we may improve queries in the future:
- We are hitting vols API to get dataNFTs each time we're opening the data explorer age. We may be able to improve this by listening for events/caching queries, and optimizing this further.

Issue #206, opened 2022-09-05
Expand UX/UI to support claiming passive rewards


idiom-bytes:
Users will be able to earn rewards in 3 ways:
![download (40).png](https://images.zenhubusercontent.com/601eeb154030993736fc619e/b6655bca-4889-4d15-a568-8258e42780a2)

Right now, our FE supports the last implementation. These rewards will be distributed via the following streams:
```
Reward #1 => OPC Revenues => FeeDistributor.vy
Reward #2-A => DF Rewards => FeeDistributor.vy
Reward #2-B => DF Rewards => DFRewards.sol
```

Right now, our UX/UI has been developed to support all requirements for Rewards #2-B.
Support for Reward#1 & Reward#2-A still needs to be developed via the front end.

### Build a claim portal for additional rewards inside FeeDistributor.vy
- [x] Update FE to show the estimated rewards that are coming through the FeeDistributor.vy contract
- [x] Implement FE + contract integration logic that enables users to claim rewards via the FeeDistributor.vy contract
- [x] Update Claim Portal to include a summary fo all rewards inside of DFRewards.sol + FeeDistributor.vy
- [x] Each claims interface should display it's own rewards (i.e. DFRewards vs. FeeDistributor rewards)

Note: There will be no way to distinguish OPC rewards from 50% of DF rewards. These numbers will be blended together as-is.

KatunaNorbert:
Why can't we just add a method to the veFeeDistributor contract to return the claimable amount?

idiom-bytes:
Continued the work on here
https://github.com/oceanprotocol/df-web/pull/223

idiom-bytes:
DF Rewards
- Got nan problem when calculating rewards, to create rewardsperlp-OCEAN.csv
- Created a manual rewardsperlp-OCEAN.csv file
- Fixed problems and ran the csv through the dispense function so I could dispense to contract successfully
- TODO: Finish visualizing + claiming rewards

VE Rewards:
- Tried to manually distribute according to PR => https://github.com/oceanprotocol/df-py/pull/286/files
- Added functionality to dispense passive rewards => https://github.com/oceanprotocol/df-py/pull/271
- Getting crashes/problems when trying to run 
```
fee_distributor.checkpoint_token()
fee_distributor.checkpoint_total_supply()
```

idiom-bytes:
Viewing rewards & claiming them is now working.
- I was able to dispense rewards via DFRewards & FeeDistributor
[...]

PR #231, opened 2022-09-15
Finalize claimable rewards

idiom-bytes:
Adjusting claimable rewards and simplifying them so user can claim them.
- I was able to dispense rewards via DFRewards contract
- I was unable to dispense rewards via the FeeDistibutor. [Please see comments inside task 260](https://github.com/oceanprotocol/df-web/issues/206)

### DIFF
--- a/src/components/claim/ClaimPortal.svelte
+++ b/src/components/claim/ClaimPortal.svelte
@@ -13,6 +13,7 @@
     rewards,
     veClaimables,
     dfClaimables,
+    getDFRewards,
     updateAllClaimables,
   } from "../../stores/airdrops";
   import { getRewardsFeeEstimate } from "../../utils/feeEstimate";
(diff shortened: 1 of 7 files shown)


idiom-bytes:
To complete the passive rewards (FeeDistributor.vy) I believe we might also need to hook it up to subgraph, as recommended in the contract. 
https://github.com/oceanprotocol/df-py/blob/4e8752abbf0d5a7050096d4ada0899d83985e1f0/contracts/ve/FeeDistributor.vy#L301

Right now, I'm still seeing a balance via the veFeeEstimate, however when I FeeDistributor.claim(), I obtain nothing. I believe this is because of how FeeDistributor moves my checkpoint.

ss_claimable_remainder

Ticket
https://github.com/oceanprotocol/df-web/issues/206

PR #232, opened 2022-09-15
Disable locked amount input if lock period is over

### LINKED ISSUE #215: Disable ocean amount to lock if lock period is over
Should we allow user to increase ocean locked amount if lock period has ended, If so we should update the logic for multiplier and voting power calculation?

KatunaNorbert:
Fixes #215.

### DIFF
--- a/src/components/veocean/LockOcean.svelte
+++ b/src/components/veocean/LockOcean.svelte
@@ -133,7 +133,8 @@
           min={$lockedOceanAmount ? 0 : 1}
           max={parseFloat(getOceanBalance($connectedChainId)).toFixed(3)}
           error={$errors.amount}
-          disabled={getOceanBalance($connectedChainId) <= 0}
+          disabled={getOceanBalance($connectedChainId) <= 0 ||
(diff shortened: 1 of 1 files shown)


src/components/veocean/LockOcean.svelte
```diff
@@ -133,7 +133,8 @@
           min={$lockedOceanAmount ? 0 : 1}
           max={parseFloat(getOceanBalance($connectedChainId)).toFixed(3)}
           error={$errors.amount}
-          disabled={getOceanBalance($connectedChainId) <= 0}
+          disabled={getOceanBalance($connectedChainId) <= 0 ||
+            new Date() > $oceanUnlockDate}
```
idiom-bytes, line 137: My only concern is that once a lock ends, there will always be an old $oceanUnlockDate, and this will never reset properly.

I believe we should just use the same checks that the contract does when trying to create lock/extend lock. I.E. Wrap functionality from the contract into a helper function, and call that.

Screenshot from 2022-09-15 12-45-25

PR #234, opened 2022-09-15
Fix lock ocean time period button click

### LINKED ISSUE #216: veOCEAN "Approve" should only be enabled if both amount + date is correct. Choosing date from shortcut should always choose a THU.
### Problem
I expect the Approval + Lock to be a clean action, but am blocked from completing it.

### How to reproduce
1. I choose a correct amount and then pick 6 months
2. I am approved for the amount I chose
3. I now receive an error that the date is incorrect
4. Choosing a 1week/6month/etc time frame, doesn't actually work (I'm blocked from locking). I need to open the calendar, and choose the nearest thursday.

### Desired States
Disabled - user amount is 0 || date is wrong
Enabled => Approve - user amount is > 0 && date is correct
Enabled => Lock - user amount is <= approved && date is correct

### DoD
- [ ] The user should have to pick a correct amount + good date, before enabling "Approve"
- [ ] Chosing 1 week, 6 month, etc... from the shortcuts doesn't actually choose a valid date. The user should not need to open the calendar.
- [ ] Approving your amount should automatically put you in a position to Lock successfully.

KatunaNorbert:
Fixes #216.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -9,6 +9,7 @@
     userAddress,
     connectWalletFromLocalStorage,
     selectedNetworks,
+    web3Provider,
   } from "./stores/web3";
   import { Router, Route } from "svelte-navigator";
   import WalletConnectModal from "./components/common/WalletConnectModal.svelte";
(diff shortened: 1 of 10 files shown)


idiom-bytes:
Tested locally, looks good.

My assets are now gone (maybe they were originally in purgatory, who knows), we might want a reset button.
I'm going to reset by hand/continue for now.

Screenshot from 2022-09-16 09-12-13

Issue #250, opened 2022-09-22
Cleanup Contract & Network Configuration - Make it easy to switch between envs


idiom-bytes:
### Problem
Right now switching between Rinkeby and local barge/ganache configurations is heavy.
In order to test longer-term functionality inside df-web we'll need to more easily reconfigure the app to run on ganache.

### DoD
- [ ] dfweb runs on top of ganache by simply changing the `VE_SUPPORTED_CHAINID` in .env 
- [ ] All contract configuration from .env has been removed
- [ ] Contract lookup code in app consumes from address.json
- [ ] Df-Web https://github.com/oceanprotocol/df-web/issues/285

idiom-bytes:
Solving for requirements in [228](https://github.com/oceanprotocol/df-web/issues/228), in order to more easily test long-term functionality.

PR #251, opened 2022-09-22
Cleaned up contract lookup logic + implemented address.json

idiom-bytes:
Addresses DoD of https://github.com/oceanprotocol/df-web/issues/250


KatunaNorbert (changes requested):
I have some issue when testing this PR, first there were some import issues which I fixed and push the changes, then now I'm getting some error when trying to retrieve Ocean balance due to chainId being undefined so it can not get the token address

KatunaNorbert (approved):

idiom-bytes:
Hi Norbert, yes this was still WIP.
I meant to mark it as a draft but it was EoD so I overlooked it.

Thank you for looking into it and taking it to completion.

PR #269, opened 2022-09-23
Add checkbox for user agreement before locking

### LINKED ISSUE #259: Add checkbox for user agreenment on locking tokens
add a checkbox under lock button that user has to agree with and check in order to create the lock

KatunaNorbert:
Fixes #259.


KatunaNorbert:
@idiom-bytes I reviewed your changes and spotted some issues:
-Token approval was disabled even if I checked the allowance checkbox. I'm also not sure If user should check the agreement for approving tokens.
-You need to update both token balance and lock end date in order for the lock button to be disabled which is wrong. The button should be disabled and allow you to update either token amount or lock end date, of both of them.

I fixed both of them, If looks good for you feel free to merge

idiom-bytes:
Just reviewed everything and it looks great.

Merging PR.

PR #270, opened 2022-09-23
Refetch values fix after withdraw and lock tokens

KatunaNorbert:
Changes proposed in this PR:

- Refetches all parameters after lock and withdraw
- Replaces balanceOfAt with balanceOf for reducing discordance between provider and contract blockheight


src/components/veocean/WithdrawOcean.svelte
```diff
@@ -29,6 +34,16 @@
       async () => {
         loading = false;
         await addUserOceanBalanceToBalances($connectedChainId);
+        const newVeOceansWithDelegations =
+          await getUserVotingPowerWithDelegations($userAddress);
```
idiom-bytes, line 38: Although it's good that we're adding this, we may still want to disable voting power from the dashboard.

I.E. Users won't be able to delegate via the FE for the next little while.

idiom-bytes (approved):
Looks good, approving it!

src/components/veocean/WithdrawOcean.svelte
KatunaNorbert: Yes, remember that, let's handle it into another PR since we have to remove it from multiple places.
For the completion of this PR let's keep it like this to verify that everything is updating correctly.

idiom-bytes: 100%

idiom-bytes:
Reviewed code and tested in Ganache. 
- Updated state variables & logic
- Fixed withdraw button title & functional logic so it displays the right title and works as expected
- Fixed logical issues and cleaned up some of the withdraw code so it can work on ganache+rinkeby.
- Manipulated chain timeline and verified withdraw works on Ganache.
- Code runs without errors against ganache/rinkeby.

Issue is ready to merge.
Will continue to verify other PRs before updating staging.

Screenshot from 2022-09-23 14-46-04

KatunaNorbert:
Hi @idiom-bytes, thanks for the updates.
I just retested your changes, I have done some changes and fixes:
- `blockTimestamp` was in seconds and `oceanUnlockDate` in milliseconds which was braking the disable withdraw button logic. I converted the `blockTimestamp` to milliseconds.
- The text in my opinion when withdraw button is enabled is better to stay same as when is locked 'Withdraw all locked' so user knows how much he will withdraw
- Removed commented console logs

I won't merge it and wait for an approval on new changes, If they look good to you feel free to merge.

PR #272, opened 2022-09-23
Fix max date and ~4years period calculations

### LINKED ISSUE #257: ~4years button exceeds by 1 week the max unlock time
To reproduce, current day Friday 23 Sept, max lock end date set to 24 Sept 2026

KatunaNorbert:
Fixes #257.


idiom-bytes:
K, I think I finally got the brunt of the datepicker resolved.
What a pain in the ass! :sweat_smile: 

There are still a couple of things that I do not like but the selection, and date locking seems to be bang on.
- Datepicker shows a weird format MM-DD-YYYY vs. DD-MM-YYYY
- Having a wrong date (selectedDate < unlockDate) isn't bound to the Approve/Lock/Update button.
- I might have overlooked a couple of calls to utc, am just triple checking this detail.

Going forward with this project if we're playing with dates:
- Let's use moment.js
- Let's use moment().utc() everywhere

@KatunaNorbert, if you could test this tomorrow, merge, and deploy, that would be awesome.

I tested things quite a bit, expanded the tests in test_functions.js, and it all seems to be working relatively flawless (except for a few things I mentioned above).

If you notice an issue, try to figure out if you can reproduce it and document it in here.
Hopefully we can get this update in place tomorrow.

Cheers!

idiom-bytes:
We reviewed this and are confident that these improvements help us get closer to the target quality level.
We have also created additional, more granular tasks, so we can take what has good effort:impact ratios.

Merging & deploying!

PR #273, opened 2022-09-24
Disable lock buttons if token amount input exceeds token balance

### LINKED ISSUE #258: Limit approval button inside lock ocean screen to max tokens inside user balance

KatunaNorbert:
Fixes #258.


idiom-bytes:
Merged latest, fixed conflict, and tested.
LGTM! Merging.

PR #274, opened 2022-09-24
Fix update veOCEAN balance when switching network

### LINKED ISSUE #265: Data is all over the place when not connected to enforced network
I get this when currently being connected to mainnet:

Screen Shot 2022-09-23 at 14 39 52

If we get the balance, we should also get the other numbers. If as an anti-pattern we absolutely require people being connected to the network we want then block the UI. Or clear all numbers.

Common solution of course is that user's connected network should only matter once they initiate transactions, all data and numbers should not require a specific user network

KatunaNorbert:
Fixes #265.


idiom-bytes:
Tested, looks good.
There is a load that gets triggered which might be not needed, but we can address it later.

PR #277, opened 2022-09-24
Disable assets table when adding/removing allocations

### LINKED ISSUE #276: Disable datasets sorting by allocated amount when changing allocations
When adding or removing allocations to an asset, do not sort the table in order to not make the user confused

KatunaNorbert:
Fixes #276.

### DIFF
--- a/src/utils/data.js
+++ b/src/utils/data.js
@@ -1,8 +1,13 @@
 export let filterDataByUserAllocation = (datasets, allocations) => {
     let filteredDatasets = []
     datasets.forEach((dataset) => {
+        if(dataset.myallocation > 0){
+            filteredDatasets.push(dataset)
+            return
+        }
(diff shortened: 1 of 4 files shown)


idiom-bytes:
This looks great, merging.

src/utils/data.js
```diff
@@ -1,8 +1,13 @@
 export let filterDataByUserAllocation = (datasets, allocations) => {
     let filteredDatasets = []
     datasets.forEach((dataset) => {
+        if(dataset.myallocation > 0){
+            filteredDatasets.push(dataset)
+            return
+        }
         allocations.forEach((allocation) => {
             if(allocation.nftAddress === dataset.nftaddress){
+                console.log(dataset)
```
idiom-bytes, line 10: I'm seeing some console logs but it seems useful.
Let's leave this on here for now.

PR #309, opened 2022-09-29
Implementing all checks and hooks for claiming rewards.

idiom-bytes:
Preparing to merge these updates again.
- Integrated veFeeEstimator.vy contract + updates to DFRewards.sol
- Tested distributing + claiming rewards from both contracts
- Tested FE by making sure that when there isn't something claimable, that the buttons remain disabled
- Tested by making sure that claim amounts are updated after the user claims successfully


idiom-bytes:
@KatunaNorbert can you please give this a review?
I have tested this locally, thoroughly, and I believe it's all working.
I pushed it early and reverted it because I did not quite have a review.
I'd like to merge this.

PR #310, opened 2022-09-30
Clear cache on new version

### LINKED ISSUE #263: Cache invalidation is missing
Whenever new version is deployed it seems there is no cache busting or anything going on. Not even manually reloading gives me new versions. In one browser tab I still have the version from 2 days ago and no easy way of getting latest version except for nuking all my browser's cache. This is not feasible and will lead to crazy amount of support requests

KatunaNorbert:
Fixes #263.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "carbon-components-svelte": "^0.63.8",
         "carbon-icons-svelte": "^11.0.1",
         "ethers": "^5.6.5",
+        "rollup-plugin-copy": "^3.4.0",
         "sirv-cli": "^2.0.0",
         "svelte-apollo": "^0.5.0",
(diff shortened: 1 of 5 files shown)


idiom-bytes:
Looks good. I tested both `npm run dev` and the docker container to make sure that the bootstrap/launcher is working correctly while connecting to it locally.

I also reviewed a few things more in detail (like html js loaders, rollup.config.js) and tried to understand a bit more of what's going on.

@kremalicious could you maybe take a quick look and provide a review for this?
I'm not very strong with rollup/package.json/overall js bootstrap/config system.

idiom-bytes (approved):
Reviewed, tested locally, and read more about the different configurations. It LGTM and am confident in merging, however I don't understand it fully so I asked kremalicious for a review.

PR #314, opened 2022-10-01
Clear cache on new version + Hot Reload works

idiom-bytes:
@KatunaNorbert let's use this PR to continue working on the clear cache.
Related to issue [#263](https://github.com/oceanprotocol/df-web/issues/263)

Issue:
- When using `npm run dev` with `NODE_ENV` not set (not prod), svelte hot reload breaks.
- When reload happens, css/styling is gone, page breaks.


idiom-bytes:
Updated branch with latest and tested again.
Merging into main.

PR #323, opened 2022-10-04
Verified claiming is working and fixed some peripheral issues

### LINKED ISSUE #247: Claiming VE Rewards - FeeDistributor.vy & FeeEstimator.vy - Everything points to user being able to claim, but is unable to do so
### Problem Description
Everything is pointing to the user being able to claim, the user has never claimed before, but is unable to do so either from the UI or calling the contracts directly.

Account in question => 0xe2DD09d719Da89e5a3D0F2549c7E24566e947260
PR w/ changes => https://github.com/oceanprotocol/df-web/pull/249

Image 1
- FeeEstimator is reporting that the user is in epoch 0, and there are amounts to be claimed.
- The FE is reflecting these values
- The user in this instance has tried to claim the rewards, has succeeded, but their balance remains the same (225.62026 OCEAN)

Image 2
- FeeEstimator in brownie is confirming the user has amounts to claim
- Transaction completed successfully
https://rinkeby.etherscan.io/tx/0x68e990b1cec4268df1365b2a4469f04c3cabd48d714ecd6fc9646be9ae965666
- User balance did not move, and FeeEstimator.vy still reports there is an amount to claim
- User epoch and value does not move after tx completes successfully

### DoD:
- [ ] Contract/calculations reports amount that the user is able to claim
- [ ] User is able to correctly claim rewards as reported

idiom-bytes:
Fixes #247


idiom-bytes:
I'm pushing this as they are minor updates, but fix issues regarding contract addresses.
Will deploy to staging.

PR #326, opened 2022-10-05
Add 100 columns for page inside data explorer

### LINKED ISSUE #306: Expand Data Explorer to 100 rows.
Right now there is only 20 items per page, increase the number of items to 100.

We can then in the near future/someone else, add a drop down with a # of rows you want to see.

As @kremalicious shared, a streaming list can result in all sorts of issues so let's take this approach first.

KatunaNorbert:
Fixes #306.

### DIFF
--- a/src/components/common/Table.svelte
+++ b/src/components/common/Table.svelte
@@ -51,7 +51,7 @@
   let loading = false;
 
   let columns = {};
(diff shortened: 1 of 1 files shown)


idiom-bytes:
LGTM! LG!

Issue #227, opened 2022-09-13
Claim Portal - Epoch Summary component


idiom-bytes:
DONE: Date - Always Thursday, hard coded.

Improve the countdown by adding additional metrics regarding the epoch, rewards, the epoch index, etc..

#### Reward Streams
Estimated DF Rewards -> 50% of weekly rewards
Estimated VE Rewards -> 50% of weekly rewards
OPC Ve Rewrds -> Whatever is coming from the OPC/Fee contract

### DoD:
- [x] Implement Epoch Summary card or component.
- [x] Reads from contract or config to tell user the estimated rewards for the week for passive & active rewards.
- [x] Provides user with forward looking reward info about the epoch

PR #331, opened 2022-10-12
Implemented epochs data + function so we can serve this data

### LINKED ISSUE #311: Voting period is vague
As a user on the "Data Explorer" screen there are only vague timeframes mentioned, making it impossible to know when to start or reassign allocations to maximize rewards:

> ...starting from Thursday, as soon as Data Farming goes live.

What Thursday? Which time? Which timezone? "as soon as Data Farming goes live", so when is that?

Please simply tell users when voting periods starts and ends on this screen. Example Balancer:

Screen Shot 2022-09-30 at 13 40 06

idiom-bytes:
Fixes #311 #227
- Estimates for each round are not available until funding is distributed, so the only way to provide this is with a data source
- Implemented an epochs.json file that enables us to configure our rounds
- Right now, there is no other source of truth to expose future round data

### DIFF
--- a/src/components/claim/ClaimPortal.svelte
+++ b/src/components/claim/ClaimPortal.svelte
@@ -15,8 +15,8 @@
   } from "../../stores/airdrops";
   import { getRewardsFeeEstimate } from "../../utils/feeEstimate";
   import { getVeOceanBalance } from "../../utils/ve";
-  import Countdown from "../common/CountDown.svelte";
   import { getAddressByChainIdKey } from "../../utils/address/address";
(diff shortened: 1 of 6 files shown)


KatunaNorbert:
@idiom-bytes Looks like the main point targeted by this PR description was already merged into main(file with the funding objects), and what is left are just design updates which would be overwrite by #341.

My suggestion is to just drop this PR

idiom-bytes:
We're not gonna push these, and are going to look into the latest from @KatunaNorbert

PR #340, opened 2022-10-18
Issue332 epoch history component

idiom-bytes:
Added more functionality around incentive history, and created a component so we can show users historical data.
Will continue to improve on these components so the reward page is more fleshed out.


KatunaNorbert:
Hi @idiom-bytes, good job here!
I did some tweaks like:
-date format for start date displayed as 'DD-MM-YYYY' to be consistent with the format we display in other places in the app
-table design updates to make it more consistent with the other table
-removed console log

I think we can merge this, looks good.

We should improve the way we use tables to be more consistent trough app now that we use multiple table, but will create an issue and we can iterate over there.

KatunaNorbert (approved):

PR #352, opened 2022-10-27
Calculate lock multiplier if lock amount field is 0

### LINKED ISSUE #296: Lock Multiplier could work without having the Ocean Amount set to > 0
If the user chooses a date range to lock, display the correct % Multiplier.

KatunaNorbert:
Fixes #296.


idiom-bytes:
LGTM! Merging.

PR #354, opened 2022-10-27
Enable to claim active rewards if user has unset allocations

### LINKED ISSUE #353: Active Rewards only visible if all allocation is allocated
The claim button for active rewards is displayed only if user has all his allocation allocated to datasets.
Same issue for rewards amount.

KatunaNorbert:
Fixes #353.


idiom-bytes:
Tested/verified.

PR #359, opened 2022-10-28
Tightening up the number input.

### LINKED ISSUE #358: Lock NaNs, negative numbers, etc... from amount input
- [x] It's still possible to enter negative numbers, get NaNs, and other bad states. 
- [x] Yup validates, but it doesn't limit the value inside the number input, so I updated the logic around the amount input. I left the ability for the user to input more than max, because there is a max button to reset it. This enables the component to still be used like a calculator, going beyond what the user may currently have in their balance. CTA/Lock remains disabled until OCEAN amount is valid.

idiom-bytes:
Fixes #358

### DIFF
--- a/src/components/veocean/LockOcean.svelte
+++ b/src/components/veocean/LockOcean.svelte
@@ -195,9 +195,7 @@
     let _amount = $form.amount;
     _amount = _amount == null ? 0 : parseInt(_amount);
     _amount = _amount < 0 ? 0 : parseInt(_amount);
-    
-    // Do not enforce max balance, let user play
-    // _amount = _amount > oceanBalance ? oceanBalance : parseInt(_amount);
+    _amount = _amount > oceanBalance ? oceanBalance : parseInt(_amount);
 
     $form.amount = _amount;
   }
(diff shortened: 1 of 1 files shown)


src/components/veocean/LockOcean.svelte
```diff
@@ -190,6 +190,17 @@
...
 
   $: calculatedMultiplier, $form.unlockDate, updateMultiplier();
+
+  const updateFormAmount = () => {
+    let _amount = $form.amount;
+    _amount = _amount == null ? 0 : parseInt(_amount);
+    _amount = _amount < 0 ? 0 : parseInt(_amount);
+    
+    // Do not enforce max balance, let user play
+    // _amount = _amount > oceanBalance ? oceanBalance : parseInt(_amount);
```
KatunaNorbert, line 200: I agree with not enforcing the amount to be less than the max amount inside input as long as we have an error message for that case. Right now there isn't any message to signal the user that he needs to change the amount in order to be allowed to continue the lock process.
Unless we have a clear message for when user exceeds the max amount I suggest we enable the enforcing.

idiom-bytes: Ok, then let's enforce. I would rather keep it minimal and avoid error messages. You can only select correct values.

PR #361, opened 2022-10-28
Add dynamic host

### LINKED ISSUE #360: Dynamic app server host
Set the server host dynamically depending on running app inside or outside a docker container

KatunaNorbert:
Fixes #360.


idiom-bytes (approved):
Thanks for addressing this, looks great. 
It will help a lot with ops as I keep forgetting to adjust the host.

PR #367, opened 2022-11-01
increase font size

### LINKED ISSUE #279: Tiny font sizes
Lots of important text is set in a computed 9.6px, which is simply too small to be readable by majority of users. Please make sure all text sizes across the whole app are at least a computed 14px by default, ideally 16px. This then also prevents `input` zooming on iOS when focused.

Like, multiple UI elements in this screenshot are 9.6px small, like the rights waiving checkbox text, or the time frame buttons which do not look like buttons:

Screen Shot 2022-09-26 at 13 31 52

KatunaNorbert:
Fixes #279.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
-    "start": "sirv public --single --port 4005"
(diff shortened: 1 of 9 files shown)


idiom-bytes (approved):
Reviewed & tested. LGTM. Merging.

PR #371, opened 2022-11-02
Add steps to simplify approve and lock

### LINKED ISSUE #305: Lock Ocean UX - Update component to simplify Approve/Lock step, and address all possible UX flows.
#### Candidate Solutions

**Approval Step:**
Candidate A: Infinite approval, similar to Uniswap.
Candidate B: Individual approval, user must approve every single amount.

**Lock Step:**
Candidate A: No popup, everything is verified via metamask
Candidate B: Extra popup, asks user to confirm their actions. << Similar to sushiswap/uniswap when you have infinite approval

#### UX Arcs
There are 3 different locking flows the user can experience. We can simplify these by 
- Creating new lock
- Updating lock (amount + time)
- Updating lock (time only)

Please refer to sushiswap UX & curve.fi so we can identify how we can improve this.

### DoD:
- [ ] We have 1 (or 3, 1 for each) multi-step component that reflects all of these flows and are contextual to the user actions
- [ ] Substitute the 1 button, for this multi-step flow

KatunaNorbert:
Fixes #305.


src/components/veocean/LockOcean.svelte
```diff
@@ -170,6 +181,21 @@
     updateLockButtonText = getUpdateLockButtonText();
   }
 
+  $: if (tokenApproved !== undefined) {
+    if (tokenApproved) {
+      currentStep = 1;
```
idiom-bytes, line 186: I have updated various things under branch => `issue305-review-approve-and-lock`
- Checkbox above the approve button
- Step component at bottom of card
- User should always start at step 0

There is some work to be done here.
- If amount passed into isCurrencyAllowed(amount) == 0, it should return false, not true by default. This should force the user to always start at step 0, not at step 1.
- After the user locks, they should be able to resume from step 1 as long as amount <= 0
- If amount > 0, then step = 0

I spent some time trying to address the details here, inside of the branch.
Please continue to merge the other PRs as appropriate.
Cheers!

idiom-bytes (changes requested):
I believe there are some adjustments to improve the UX, and that the step component isn't quite there yet. 

I.E. The user often starts at step 1, when they should start at step 0.
- If no lock and amount is 0, then step is 0
- If lock and amount is 0, then step is 1
- if lock and amount > 0, then step is 0
- I think the checkbox might be moved before the Update Lock button.
- I don't think the "Update Lock" button description should change to "Update Lock End Date". Again, I feel we're trying to make things smart, but just adding complexity. Another example is making the "Lock Ocean" card change titles.

Generally, I feel that the number of steps should change, rather than having the step where the user is at change. As described in issue #305. Based on the current implementation, I'm not suggesting we change the step component... just tighten up a bit of how the user flows through the steps.
https://github.com/oceanprotocol/df-web/issues/305

Thank you!

KatunaNorbert:
I can agree with moving the steps component at the bottom and agreement checkbox above the button but I'm not sure about showing the first step by default.
It doesn't really make sense to show 2nd step at 0 balance when there is a lock and 1st when there isn't. It should be the same outcome since it doesn't rely on have a lock.

idiom-bytes (approved):

idiom-bytes:
We went over the code I provided in https://github.com/oceanprotocol/df-web/pull/397 and merged that into this PR.
This now has everything required to deliver the step component, and make it play well w/ the state of the user.

Reviewed implementation w/ Katuna

idiom-bytes:
Tested/reviewed. Merging.

PR #386, opened 2022-11-08
Update available allocations section

### LINKED ISSUE #378: Update allocation tooltip FTUE
### Problem:
- Right now when viewing the data explorer, it's not quite obvious that the user must allocate

### Candidate solution:
- Allocation tooltip could be better visualized to show the user that they have not yet completed the action, such as using an alert/red tooltip, changing the text and color only after user completed action.
- Change the amount shown to be a fraction (0/100). Using 100% which doesn't quite map to what's happening under the hood. Showing zero to the user, better communicates that they have not yet completed it.

### On the alert (optional, use your discretion when delivering this)
- This may be a red exclamation mark, or a font color, or a background color.
- However it should change afterwards

### DoD:
- [ ] Alert should only show if the user has logged into wallet, but has no allocations. There may be a different/unique description telling the user they have not yet locked.
- [ ] When the user completes the first allocation, this "you have not yet locked" goes away, and shows a generic description..
- [ ] When first arriving at the data explorer page, it's obvious to the user that they must allocate if they have not done so yet

KatunaNorbert:
Fixes #378.


idiom-bytes:
Worked really well. I think we can tighten the UX a good amount here.
The top part is very verbose compared to the rest of the app.
Make it a H1 kinda thing w/ the text above as a description. Or, <h1>Data Explorer<h1> and shove everything else into that tooltip. I.E. Increase context/visual feedback of (!) vs. adding more text in the open.
Screenshot from 2022-11-15 10-34-52

I got my allocations set, but then they disappeared... i.e. 100/100, but they are not visible.
Screenshot from 2022-11-15 10-50-51

 Didn't debug, instead, I tried to reset my allocations, and it Errored - Stating "You have no allocations set."
Screenshot from 2022-11-15 10-51-02

KatunaNorbert:
**Issue 1**
I retested and I don't think there is an issue with what is displayed. You should have lock If the Reset Button and percentage shows that.
Regarding your error, have you clicked on Update or Reset? If you clicked on update, error makes sense because you haven't changed your allocations distribution. If you clicked on reset, I guess issue is related to button action, although it shouldn't be the case to show that error If you click reset.


**Issue 2**
We could move text to tooltip, the only concern is that we should keep the current description but then also add the new description which will make a long text to read. Also user has to hover the tooltip in order to see the message.

idiom-bytes:
```Issue 2
We could move text to tooltip, the only concern is that we should keep the current description but then also add the new description which will make a long text to read. Also user has to hover the tooltip in order to see the message.```

I believe that if the UX-UI is self-explanatory enough, we should not need any long-form text.
I would argue that the ux is getting there.

In games, we try to have as little text and reading in front of the user as possible.
I believe we would benefit from adopting that here.
It also helps make other text stand out more.

0 sentences. 
![Mirage-10](https://user-images.githubusercontent.com/69865342/202269659-289a3215-affc-4880-9de3-4ec23a304bc6.png)

idiom-bytes:
P.S. I don't believe the yellow tooltip is enough at the moment. Did you checkout carbon components icons for something that might look a bit better?

idiom-bytes:
I iterated on the component/element, and pushed a PR that should update this branch/pr.
https://github.com/oceanprotocol/df-web/pull/393

idiom-bytes:
Tested & reviewed. Looking great.

PR #389, opened 2022-11-11
Improve asset total allocated value

### LINKED ISSUE #388: Improve asset total allocations display
The current TotalAllocated column contains the sampled allocation, which means that if user moves his allocations from asset A to asset B, this will happen gradually due to the sampling process, and the entire allocations will be moved over only at the beginning of the next round. 
Problem is that users expect to see the allocations moved over to the new asset right away, or as soon as possible.

This creates confusion for the users, although this is exactly what it's going on in the perspective of rewards calculation and it's not an issue.
To make this more clear, create 2 separate table column as following:
1.Rename the current column which contains the sampled total allocations and add more information inside a tooltip.
2.Add a new column which contains the not sampled total allocation. This will confirm the user that his allocation have moved to the new asset and users can also have a better overview of where the allocations are without having to wait until the round has ended.

KatunaNorbert:
Fixes #388.


idiom-bytes:
@idiom-bytes update tooltip descriptions and column names

PR #392, opened 2022-11-16
Rewards Page - Passive & Active APYs - Reward Cards & Overview Cards

### LINKED ISSUE #391: Display rewards generic(not user specific) APY
Display the not user specific APY inside df-web.

-APY for passive rewards should be calculated on df-web side.
-APY for active rewards should be fetched from df-sql.

KatunaNorbert:
Fixes #391.


idiom-bytes:
Ready for review & potentially merge

idiom-bytes:
Yes. Going to merge this too. Pink text makes it pop :tada:

idiom-bytes:
I reviewed the PR and although great in general, there are some issues with it.
- Primarily, the passive apy calc is wrong.

APY Calculations are wrong, reporting any apy of (7.14 e^104).
Screenshot from 2022-11-17 16-19-34

APY Calculations in `rewards.js` is where the issue is likely occuring
Screenshot from 2022-11-17 16-19-45

Need to review/cleanup APY calculation

KatunaNorbert:
I think calculation is fine. The problem is that `veSupply` on Goerli network is very low `264.97` that's why APY is that high.

KatunaNorbert:
As a fallback I'm suggesting to have a threshold, maybe 10000 and If APY is greater than that display 'over 10000'.
We won't have this issue on production and for test networks It it's going to display a better value which doesn't impose error.

idiom-bytes:
```I think calculation is fine. The problem is that veSupply on Goerli network is very low 264.97 that's why APY is that high.```
Fair, but we can't rely on production settings for verification.

I think we should work out a way to validate these calculations in the app via an envvar, or to write some tests in js (like I did with other functionality) that validates the implementation on the client and enables me to verify that it's correct.

```As a fallback I'm suggesting to have a threshold, maybe 10000 and If APY is greater than that display 'over 10000'.```
I don't think we should do this.

Again, just a vanilla JS test in the test folder, would help me validate it and give me confidence that the functionality works.
If we do something for dev, then we should make it easy/accurate.
I.E. Reduce the reward amounts in dev/staging so we can get a functional number to work with.

**Finally:** I'm going to merge this into staging so we can move it forward.
I do think it would be good to setup tests for any of this functionality that can be easily verified in js.

PR #393, opened 2022-11-16
Cleaned up tooltips, icons, colors, added multiple state for tooltip,…

idiom-bytes:
- Eliminated all that text at the top
- Created a clearer title for the data explorer so we can more easily refer to it.
- Updated the tooltip component so it has:
- 3 states that change colors & icon to better guide the user
- Multiple tooltip descriptions, so it changes and instructs the user accordingly
- Embedded all the onboarding logic into the tooltip. Tooltip has 3 states.
1. Alert - Red/Filled - User has no lock - Instructs him to get lock
2. Warning - Yellow/Filled - User has lock but no allocation - Instructs him how to allocate
3. Normal - Black/NoFill - User has allocated - Maintains same instruction

datafarming_no_lock
datafarming_has_lock
datafarming_has_allocation
Screenshot from 2022-11-16 15-09-27


idiom-bytes:
- Updated title in rewards page.
- Changed REWARDS => Rewards
- Other minor details

idiom-bytes:
I took a bit more time to iterate on the copy of the Reward cards to:
- Make them read **very similarly** to each other, while **highlighting** their differences
- Used the same words and concepts that are repeated in the cards, to emphasize the values and actions the user must take
- Continued this as part of "writing tooltips is not easy without a doc"

PR #397, opened 2022-11-17
Step Component Adjustments - Final touches

idiom-bytes:
I have updated various things under branch => issue305-simplify-approve-and-lock

Checkbox above the approve button
Step component at bottom of card
User should always start at step 0
There is some work to be done here.

If amount passed into isCurrencyAllowed(amount) == 0, it should return false, not true by default. This should force the user to always start at step 0, not at step 1.
After the user locks, they should be able to resume from step 1 as long as amount <= 0
If amount > 0, then step = 0
I spent some time trying to address the details here, inside of the branch.
Please continue to merge the other PRs as appropriate.
Cheers!


idiom-bytes (commented):
Reviewed w/ @KatunaNorbert, and simplified to cut remaining work as part of https://github.com/oceanprotocol/df-web/issues/398

PR #402, opened 2022-11-23
Created test to verify local apy, and updated function so we can have…

### LINKED ISSUE #401: Enforce reward numbers for ganache/goerli, so human APYs are generated
#### DoD:
- [x] Rewards on df-web are verifiable via tests
- [x] Rewards on df-web are verifiable via FE
- [x] Update epoch functionality to override rewards in ganache/staging environments, and make it comparable to veOCEAN supply

idiom-bytes:
Fixes #401


KatunaNorbert (approved):
Looks good!

PR #413, opened 2022-11-29
Extend min lock period

### LINKED ISSUE #400: Update create lock configuration - 2 week min
### Problem/Request:
Can we drop the 1 week lock time... there are so many people that are going to try it, but they are not going to earn any rewards. they need a lock of at least 2 weeks... for the first. we could add a disclaimer on the website.

```
if you lock for 1 week => you might get active rewards, but you're not gonna get any passive rewards
if we lock for 2+ weeks => then it's hard for people to mess up. there are a lot of users not getting rewards
```

Note, we should use text, language, descriptions, from the blog posts as much as possible. We can improve these, but we should take the base writings from the blog posts.

### Specifically:
"There’s a new DF round every week; in line with this, there’s a new ve distribution “epoch” every week. This affects when you can first claim rewards. Specifically, if you lock OCEAN on day x, you’ll be able to claim rewards on the first ve epoch that begins after day x+7. Put another way, from the time you lock OCEAN, you must wait at least a week, and up to two weeks, to be able to claim rewards. (This behavior is inherited from veCRV. Here’s the code. )"

### DoD:
- [x] Update descriptions on df-web to better explain the initial lock times & min time fo rewards
- [x] Update ocean lock functionality so there is a min 2 week period for the lock. We can use the existing functionality to get the thursday (as is) and then simply use the moment object it returns, to advance the moment by 1 week.

KatunaNorbert:
Fixes #400.


src/components/common/DateInput.svelte
```diff
@@ -43,8 +43,7 @@
 <div class={`container`}>
   <input
     class={`input ${
-      (value === min && moment.utc().format("YYYY-MM-DD") === min) ||
-      (value <= min && moment.utc().format("YYYY-MM-DD") <= min)
+      value < min && moment.utc().format("YYYY-MM-DD") <= min
```
idiom-bytes, line 46: I don't believe this is working, but when I went to test it, the calendar is not showing anymore.

src/components/veocean/LockOcean.svelte
```diff
@@ -91,7 +91,9 @@
     amount: 0,
     unlockDate: $oceanUnlockDate
       ? $oceanUnlockDate.format("YYYY-MM-DD")
-      : moment.utc(getThursdayDate(moment.utc())).format("YYYY-MM-DD"),
+      : moment
+          .utc(getThursdayDate(moment.utc().add(7, "days")))
+          .format("YYYY-MM-DD"),
```
idiom-bytes, line 96: These updates look correct. Thank you!

KatunaNorbert: Cool, thanks for checking

src/components/common/DateInput.svelte
idiom-bytes: Brave... but have chrome/firefox for testing.
Good point on input disabled.
Will respond again if I give it a try.

EnzoVezzaro:
I think this is not working as expected. On my end I can't select February:

Screenshot 2023-01-30 at 11 29 31

KatunaNorbert:
@EnzoVezzaro  I can't reproduce the date picker issue on my end, maybe you can give some more information about how to reproduce it.
The description was update, but was not marked as done in the DoD list from within the issue.

EnzoVezzaro:
> @EnzoVezzaro I can't reproduce the date picker issue on my end, maybe you can give some more information about how to reproduce it. The description was update, but was not marked as done in the DoD list from within the issue.

Sure, I'll try to get some more info, or maybe I can find a fix directly. I think is related to my machine / timezone. But I'll let you know once I have more info.

src/components/common/DateInput.svelte
jamiehewitt15: Yeah, I also don't see the date picker on chrome
image

jamiehewitt15: It seems to work on Firefox and Brave for me through

jamiehewitt15: Without the date picker this is very frustrating to use as it's not possible to type the date either. So the only option is to select one of the predefined options below

KatunaNorbert: Hm.. I'm using Chrome but it is visible for me

jamiehewitt15: Possibly something that was already fixed on main. I just saw you pulled from main and now the date picker is working fine for me  👍

KatunaNorbert:
Ok, @EnzoVezzaro I leave this PR on you, thanks!

src/components/common/DateInput.svelte
EnzoVezzaro: Still investigating, but on my end actually Chrome works correctly 😅 

Chrome:
Screenshot 2023-02-02 at 08 02 08

Brave:
Screenshot 2023-02-02 at 08 02 17

jamiehewitt15: Yeah, weird that the min value is different

KatunaNorbert: Well, looks like on Brave, the connected account already has a lock, so that changes the min value

EnzoVezzaro: this was caused by a plugin I had on my machine. No further discussion needed.

EnzoVezzaro:
found the issue, now seems all solved. At the end it was on my machine, I had a plugins which confused `momentjs` on different browsers and also interfered with the wallet connection. I'll go ahead an approved this PR as min lock period is working.

EnzoVezzaro (approved):

jamiehewitt15:
Why do we still have the button bellow to select 1 week if you are no longer allowed to lock for 1 week? 

image

EnzoVezzaro (approved):

PR #420, opened 2022-12-01
import walletconnect dependencies into app

### LINKED ISSUE #369: Replace WalletConnect from unpkg (June 2023)
We currently get WalletConnectProvider from unpkg which is based on an api call. This is not ideal because if the api is down we can't use WalletConnect.
We should find a different way, ideally an npm package to get the WalletConnectProvider.

kremalicious:
- fixes #369

It's as simple as that, or it should be.

Adding as PR as couldn't figure out how to get local version running, so hoping for PR preview deployments to test.

If indeed app is not setup to import either commonjs or esm packages, then most likely for this case here this rollup plugin needs to be setup:
https://github.com/rollup/plugins/tree/master/packages/commonjs


public/index.html
```diff
@@ -8,8 +8,7 @@
 					  <link rel='icon' type='image/png' href='/logo-ocean-svg.svg'>
 					  <link rel='stylesheet' href='/global.css'>
 					  <link rel='stylesheet' href='/build/bundle.css'>
-					  <script type="text/javascript" src="https://unpkg.com/@walletconnect/web3-provider"></script>
-					  <script defer src='/build/bundle.js'></script>
+					  <script defer src='/build/bundle.5237e218.js'></script>
```
kremalicious, line 11: note to self: this change with the hashed bundle.js is unintended, revert

idiom-bytes:
Hi @kremalicious I'm reviewing this PR:
1. The website starts breaking => https://df-9sf551740-oceanprotocol.vercel.app/
I believe this is the result of the problem described here: https://github.com/sveltejs/svelte/issues/6404

When @KatunaNorbert implemented WC, there were a handful of modifications needed in order to get it working.
https://github.com/oceanprotocol/df-web/commit/7a4abc28853c9e93689da7fedf5047f26c79693c

Perhaps there is something else that was overlooked?

As far as I can tell, this modification requires an `import` that may be throwing an error due to how the underlying .js code gets compiled.

kremalicious:
Most likely:

> If indeed app is not setup to import either commonjs or esm packages, then most likely for this case here this rollup plugin needs to be setup:
https://github.com/rollup/plugins/tree/master/packages/commonjs

KatunaNorbert:
I already tried to add `commonjs` to rollup plugins but not working, that's why I didn't use the walletconnect dependency.
Will try to look more into enabling support for commonjs packages in this PR.

idiom-bytes:
Converted back to draft until there is a potential solution.

kremalicious:
ok, you might want to explore migrating to v2 instead of keeping this here as a draft until walletconnect v1 stops working completely. Basic problem and (react-only) options outlined here:
https://github.com/oceanprotocol/market/issues/1828

KatunaNorbert:
I have been looking into migrating to WalletConnect v2 but, I run into the same import issue as before when trying to import WalletConnectProvider from `@walletconnect/web3-provider`. Now the issue appears when importing from `@web3modal/ethereum`.

I've been also looking for other solutions for replacing WalletConnect but didn't find one.

KatunaNorbert:
Current state of this PR:

- Replaced Rollup with Vite
There are no longer issues with importing different module types, and WalletConnect libraries can be used right now.
In terms of changes the biggest one is related to the way **env variables** are now defined and used inside the app, so for the deployment to work we have to change the way env variables are defined inside Vercel config(they should stare with `VITE_`). Will do this once PR is ready for review.

- Migrate to WalletConnectV2
I've been implementing the migration in here using the `Vanilla JS` solution which uses `wagmi` and `ethersJS client`
There are some changes on the way we read and write to the smart contracts but not too much since we were using ethersJS before.

Last thing here would be to make the mobile wallet connection trough generated QRCode work. I'm following there discord channel and turns out there are still some issues and they are preparing a new release next week for both `web3modal` and `wagmi`

EnzoVezzaro:
> channel and turns out there are still some issues and they are preparing a new release next week for both `web3modal` and `wagmi`

any news on this? is this PR still relevant?

EnzoVezzaro:
As it is, there's few things no working:

- QR is not working (try metamask)
- Zerion and Infinity gets stuck in loading

What's working:
- Metamask
- Coinbase Wallet

Tried also to update the repo to catch up with main, but there's so many conflicts I'm not sure I can get it running in a short time. Can you maybe help me here @KatunaNorbert ? I just need to have this merged with main, before I'm able to update packages for WC. I also didn't want to spend too much time running into issues you might have seen before.

EnzoVezzaro:
UPDATE: after updating WC, it seems they (WC) had remove support for Metamask, and other wallets (as discussed [here](https://github.com/WalletConnect/walletconnect-monorepo/issues/651)). 

The current state of the art is: 

- [ ] You can use Zerion to login, but we need to fix some connection issue on the logic. 

Screenshot 2023-02-24 at 09 29 31

On the market we have this login process: 

- Connect button:
Screenshot 2023-02-24 at 09 35 02

You can login with either metamask or select WC to use the other wallets like Zerion. 

- [ ] we need to add the correct information for the login modal:

DF Web:
IMG_7781

Market: 
IMG_7782
(this has more information about the connection and allows user to select the chain)

- [ ] we also need to figure out what's the impact of having Vite instead of Rollup. I noticed some changes to the overall design (larger columns and texts) that might be the consequence of the migration from rollup to vite or the marge with latest main. As well as some build problems with Vercel. 

We might want to reconsider this decision as the WC team is already[ working on a solution](https://github.com/WalletConnect/walletconnect-monorepo/issues/864) for WC with rollup, and It'll be implemente soon I hope, or work on changing completely build tool (which requires some time). 

We need to decide if this is a High priority task to start working on these points, or we need to prioritise other more important tasks.

KatunaNorbert:
@EnzoVezzaro Rollup has more downsides, especially working with bundling different module types, that's why the migration to Vite.

Some updates related to WC V2:
- They extended the deadline for switching to v2 to June
- Currently this are the mobile wallets supporting WC V2(no Metamask and Ledger):https://explorer.walletconnect.com/?version=2. I have been using Trust Wallet for testing
- There might be other complications until they fully move to v2

KatunaNorbert:
@idiom-bytes Vercel Preview deployment it's now working and WC V2 it's ready to be tested

package.json
```diff
@@ -1,57 +1,51 @@
...
   "private": true,
+  "version": "0.0.0",
+  "type": "module",
   "scripts": {
-    "build": "rollup -c",
-    "dev": "HOST=localhost rollup -c -w",
-    "start": "sirv public --single --port 4005 --host 0.0.0.0"
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite preview"
```
idiom-bytes, line 9: build times seem to have improved. great!

src/stores/web3.js
idiom-bytes, line 1: great cleanup!

src/utils/feeDistributor.js
```diff
@@ -1,66 +1,92 @@
...
   } catch (error) {
-    console.log(error)
+    console.log(error);
     throw error;
   }
-}
+};
 
-export async function claim(userAddress, signer) {
+export async function claim(userAddress) {
```
idiom-bytes, line 61: Did not test latest implementation of claim functions inside in-app/test environment, only goerli.
Will continue with testing throughout this week.

src/utils/feeEstimate.js
```diff
@@ -1,37 +1,69 @@
...
-    console.log(error)
+    console.log(error);
     throw error;
   }
-}
+};
 
-export async function claimVERewards(userAddress, signer) {
+export async function claimVERewards(userAddress) {
   try {
```
idiom-bytes, line 34: Did not test latest implementation of claim functions inside in-app/test environment, only accessible functions in local goerli. Have reviewed it and implementation looks valid. Will continue with testing app throughout this week.

idiom-bytes:
Merge the PR, then delete Counter.svelte. :raised_hands:

PR #422, opened 2022-12-01
new minimal readme

kremalicious:
started by removing everything which is not erelated to this app, 

Moved env var explanations into the place where developers need them, goal is that the default env file created after `cp .env.example .env` is fully working for local development out of the box


idiom-bytes (approved):
Looks good, will continue iterating.

Thank you for breaking up the .env.example out of the readme, i agree that's a much better pattern.

idiom-bytes (approved):
Approving again, in case it didn't register.

kremalicious:
yeah, not sure about this branch protection setup, which is set to enforce preview deployments on this repo, which is why we can't merge it for whatever reason. We clearly have a preview deployment as the Vercel comment shows

KatunaNorbert:
@kremalicious thanks for cleaning the readme.
This is strange, the PR shouldn't be blocked.

KatunaNorbert (approved):

KatunaNorbert:
I think the merge blocking was due to branch not being up to date with main branch, there is a dependency for this, but the message from github was wrong.

PR #423, opened 2022-12-01
Disable APY components

idiom-bytes:
Just wrapped them. Remove wrapper and enable after APYs calcs are stable.

### DIFF
--- a/src/components/claim/RewardOverview.svelte
+++ b/src/components/claim/RewardOverview.svelte
@@ -6,40 +6,54 @@
   import Countdown from "../common/CountDown.svelte";
 
   export let roundInfo = undefined;
-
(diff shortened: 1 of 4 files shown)


kremalicious (approved):
I mean sure, if only intention is to hide them, this will do it. But would be better to just remove the code.

Keeping unused code in a version-controlled app is usually an anti-pattern as you also push this code to all users, but then they will never use it.

graceful-coder (dismissed):

idiom-bytes:
@kremalicious i don't know why i thought it would be better that way.
i have instead removed the code, we can bring it back by going through the ticket & commits.

kremalicious (approved):

PR #424, opened 2022-12-05
base APY calculations in OCEAN

### LINKED ISSUE #418: APYs must be calculated off OCEAN, not veOCEAN. For passive rewards.
Please refer to [issue #28 in dfsql](https://github.com/oceanprotocol/df-sql/issues/28) for details on how to calculate APR.
Verify calculations are correct based on the latest definition of APR.

```
Everywhere, the correct definition of APR is:
APR = OCEAN_rewards / OCEAN_locked
```

### Web App DoD:
- [ ] passive rewards global
- [ ] passive + active rewards global 
- [ ] take the latest numbers from dfsql

KatunaNorbert:
Fixes #418.


idiom-bytes (approved):
Not sure if there are additional updates for this PR. It's still marked as a draft, but I'm going to approve it based on what I see.

idiom-bytes (approved):
LGTM!

idiom-bytes (dismissed):
Approving changes

idiom-bytes:
When the user is not connected "Your APY" should not be reported.
Screenshot from 2022-12-07 19-25-33

If the user is connected, but does not have any veOCEAN, "Your APY" should not be reported
Screenshot from 2022-12-07 19-25-48

idiom-bytes:
Tested the latest version.

When the user connects, the APYs do not look correct.

In this case, the user has no veOCEAN and the Reward Overview component, is reporting your APY is 4.4%.
In the case where the user has no veOCEAN, we should continue to only report a single number. Not "Your APY"
Screenshot from 2022-12-07 20-05-55

idiom-bytes:
We may at some point want to better visualize the progress of rewards being counted/distributed. Like... instead of just a clock counting down, we could visualize the samples being taken, and the APY for the roundbeing locked in.

Right now, the user doesn't too much context about how the epoch progresses, and how the apy can start at 0%, and increase throughout the week. I'm not recommending we fix this now... I'm just documenting the problem and sharing some thoughts so we can reflect more on it.
Screenshot from 2022-12-07 20-03-02

KatunaNorbert:
@idiom-bytes thanks for the feedback.
This is still on progress now with the changes related to user passive APY, so I'm still iterating on some of the issues.
Good catch with user not having veOCEAN, it shouldn't take the passive rewards in consideration.

idiom-bytes (approved):
Looks great!

PR #425, opened 2022-12-05
Handle no epoch config object

### LINKED ISSUE #404: Handle no epoch config for current DF round
We have an object with all past and future DF round configs.
We need to handle the scenario when there is no more config in the future for current round, so we are reminded to add them but then app doesn't crash or show some concerning values.

KatunaNorbert:
Fixes #404.


idiom-bytes (approved):
I see what you did, thank you for adding these checks. I'll update epoch.json to be more forward-looking.

PR #426, opened 2022-12-07
Improve user guidance in app

### LINKED ISSUE #346: Improve UX-Flows. Improve guiding the user.
I see a couple of spots where we can keep guiding the user towards the right places
- [x] In Rewards Page => Allocate => Take user to DF Page
- [x] In DF Page => Update Allocation => Take user to veOCEAN lock page
- [x] If user has no OCEAN/About => Documentation

any other ones?

KatunaNorbert:
Fixes #346.

### DIFF
--- a/src/components/header/Header.svelte
+++ b/src/components/header/Header.svelte
@@ -6,11 +6,6 @@
   import moment from "moment";
 
   let aboutURL = "https://docs.oceanprotocol.com/veocean-data-farming-tutorial";
-
-  $: if ($oceanUnlockDate && $oceanUnlockDate > moment()) {
-    aboutURL =
-      "https://blog.oceanprotocol.com/ocean-data-farming-series-c7922f1d0e45";
-  }
   const location = useLocation();
 </script>
 
@@ -42,9 +37,7 @@
       </li>
       <li class:active={false}>
         <a
-          href={$oceanUnlockDate && $oceanUnlockDate > moment()
-            ? "https://blog.oceanprotocol.com/ocean-data-farming-series-c7922f1d0e45"
-            : "https://docs.oceanprotocol.com/veocean-data-farming-tutorial"}
+          href={aboutURL}
           target="_blank"
           class="link">ABOUT</a
         >
--- a/src/components/header/Header.svelte
+++ b/src/components/header/Header.svelte
@@ -2,9 +2,15 @@
   import WalletConnect from "./WalletConnect.svelte";
   import NetworkSelection from "./NetworkSelection.svelte";
   import { Link, useLocation } from "svelte-navigator";
+  import { oceanUnlockDate } from "../../stores/veOcean";
+  import moment from "moment";
 
-  const aboutURL =
-    "https://blog.oceanprotocol.com/ocean-data-farming-series-c7922f1d0e45";
+  let aboutURL = "https://docs.oceanprotocol.com/veocean-data-farming-tutorial";
+
+  $: if ($oceanUnlockDate && $oceanUnlockDate > moment()) {
+    aboutURL =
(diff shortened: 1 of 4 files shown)


src/components/header/Header.svelte
```diff
@@ -2,9 +2,15 @@
...
   import NetworkSelection from "./NetworkSelection.svelte";
   import { Link, useLocation } from "svelte-navigator";
+  import { oceanUnlockDate } from "../../stores/veOcean";
+  import moment from "moment";
 
-  const aboutURL =
-    "https://blog.oceanprotocol.com/ocean-data-farming-series-c7922f1d0e45";
+  let aboutURL = "https://docs.oceanprotocol.com/veocean-data-farming-tutorial";
+
+  $: if ($oceanUnlockDate && $oceanUnlockDate > moment()) {
```
idiom-bytes, line 10: I don't think we need to be this granular with the about link.

Let's just improve the docs and update the about link to point to it.

src/components/header/Header.svelte
```diff
@@ -35,7 +41,13 @@
         <Link to="/rewards" class="link">REWARDS</Link>
       </li>
       <li class:active={false}>
-        <a href={aboutURL} target="_blank" class="link">ABOUT</a>
+        <a
+          href={$oceanUnlockDate && $oceanUnlockDate > moment()
+            ? "https://blog.oceanprotocol.com/ocean-data-farming-series-c7922f1d0e45"
+            : "https://docs.oceanprotocol.com/veocean-data-farming-tutorial"}
```
idiom-bytes, line 47: I don't think we need to be this granular with the about link.

Let's just improve the docs and update the about link to point to it.

src/components/header/Header.svelte
KatunaNorbert: agree, done

src/components/header/Header.svelte
KatunaNorbert: done

EnzoVezzaro:
Not sure it's related, but when I get to the "rewards" page, I see the "Update allocations" active and if I press without changing anything it will show a modal with an error. Can we disable this button if there's no change yet?

Screenshot 2023-01-30 at 11 59 44

After pressed:
Screenshot 2023-01-30 at 11 59 35

as for the "reset" button, it looks like disabled, but you can actually press on it. Can we give it another style? Right now looks disabled, and it's kind of confusing.

KatunaNorbert:
@EnzoVezzaro I checked the reported issues and they are valide ones, but they are outside of the cope of this PR.

- Disabling the Update Allocations button if the user doesn't change his allocation might need some more work so I suggest moving this into a separate issue.

- About the Reset Allocations button, in DF-WEB there are primary and secondary button with different styles so we can highlight more important button where there are multiple ones. I guess we could do some improvement, either make the Reset Allocation button as primary or just updating the secondary button design overall, but again I would put this into a separate issue, might need some more discussion.

EnzoVezzaro:
> @EnzoVezzaro I checked the reported issues and they are valide ones, but they are outside of the cope of this PR.
> 
> * Disabling the Update Allocations button if the user doesn't change his allocation might need some more work so I suggest moving this into a separate issue.
> * About the Reset Allocations button, in DF-WEB there are primary and secondary button with different styles so we can highlight more important button where there are multiple ones. I guess we could do some improvement, either make the Reset Allocation button as primary or just updating the secondary button design overall, but again I would put this into a separate issue, might need some more discussion.

sounds good to me 👍🏽 I had moved this to #455

EnzoVezzaro (approved):

KatunaNorbert:
cool, thanks @EnzoVezzaro

Issue #377, opened 2022-11-03
Improve headers, tooltips, explainers, and links in df-web app


idiom-bytes:
### Problem:
There are still a handful of headers, tooltips, explainers, and links that might help to improve the overall UX and may be relatively low hanging fruits.

1. There are sections where the title is within a card, and there are sections where titles are outside of cards. This leads to an inconsistent UI.
2. There are sections where we've added good explainers, but are not linking to documentation, blog posts, or deeper intuition on important concepts. Passive and active rewards as an example, should be linking to the documentation where these concepts are explained w/ their latest implementations.
![Screenshot from 2022-11-02 17-20-24.png](https://images.zenhubusercontent.com/601eeb154030993736fc619e/7c5a04be-6df5-4b2e-b62f-796ec08eebbd)

1. Another example of a title outside of a card. In this area, we could likely improve the title w/ a description.
2. Allocation tooltip could be better visualized to show the user that they have not yet completed the action, such as using an alert/red tooltip and only changing color once user completed action.
3. Slight changes to button state may help to guide user.
![Screenshot from 2022-11-02 17-23-44.png](https://images.zenhubusercontent.com/601eeb154030993736fc619e/92c95749-fb15-49c8-9c74-ac776fd4f1a4)

### DoD:
- [ ] Titles are either in-cards or outside of cards throughout the app, bringing consistency to the UI
- [ ] Passive and active rewards are properly linking to their respective sections
- [ ] Epoch tooltips inside of data explorer (7d DCV + 7d Allocations) better communicate these are sampled, and therefore the amounts may not be what users expect
- [ ] Tooltip/UI for allocation better communicates when the user has completed his initial allocation
- [ ] Move Tooltip form "Rewards Page - Historical Epoch Table - Round" to "Start Date". Right now, it seems fairly redundant.

idiom-bytes:
**[Improving round description]**
Tooltip inside of Round info is fairly redundant. We may want to move this to the "Start Date" column, or as a a text_field to describe the epoch/start date/ rounds, or just remove it and continue to point users towards documentation.

![Screenshot from 2022-11-03 08-09-36.png](https://images.zenhubusercontent.com/601eeb154030993736fc619e/9d08bbed-af91-4247-b835-d5fe284cf4c3)

Norbert: In general, it would be great to have the description of the round/data farming displayed earlier to the user.
Candidate A: Move this description into the Countdown card.
[...]

PR #429, opened 2022-12-08
Issue427 - Add tooltips to apy + improve rewards page

idiom-bytes:
- Fixed Reward Title
- Removed tootlip from Rounds, and moved it to Reward Title
- Modified card component to implement Tooltip inside of subtitle to fix issues inside of #377


idiom-bytes:
The only thing remaining, is to justify the items in the <div class="subtitle"> element.
Screenshot from 2022-12-07 19-35-20

KatunaNorbert (approved):
Looks good, just pushed some small fixes.

PR #441, opened 2022-12-19
Enable sorting on 'My allocations' column

### LINKED ISSUE #377: Improve headers, tooltips, explainers, and links in df-web app

KatunaNorbert:
Fixes #377.

### DIFF
--- a/src/stores/data.js
+++ b/src/stores/data.js
@@ -24,7 +24,7 @@ export const columnsData = [
   { key: "did", value: "DID" },
   { key: "roundallocation", value:"RoundAllocation", display: (allocated) => allocated + ' veOCEAN', tooltip: descriptions.default.tooltip_datafarming_round_allocation},
   { key: "currentallocation", value:"CurrentAllocation", display: (allocated) => allocated + ' veOCEAN', tooltip: descriptions.default.tooltip_datafarming_current_allocation},
-  { key: "myallocation", value:"MyAllocation", sort: false, tooltip: descriptions.default.tooltip_datafarming_my_allocation },
+  { key: "myallocation", value:"MyAllocation", tooltip: descriptions.default.tooltip_datafarming_my_allocation },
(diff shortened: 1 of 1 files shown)


EnzoVezzaro:
Hi @KatunaNorbert sorry, but do you know why I can't see the table on /data page?

Screenshot 2023-01-27 at 08 20 02

I might be missing something since I just got here 😅

EnzoVezzaro (approved):
LGTM

EnzoVezzaro:
Retested 👍🏽 It's working as expected.

EnzoVezzaro (approved):

PR #467, opened 2023-02-21
Display publisher and publisher rewards in DF table

### LINKED ISSUE #461: Update DF Table - Publisher Rewards
### DoD:
- [x] Add drop down with 3 filters: all datasets, assets with allocations, my published assets
- [x] Add publisher address + link to marketplace
- [ ] Ready to go. Please finish review & publish.

KatunaNorbert:
Fixes #461.


src/components/common/Table.svelte
```diff
@@ -292,13 +295,14 @@
...
-            type="checkbox"
-            label="Only data where I have allocations"
-            bind:value={showDataWithAllocations}
-          />
-        </div>
+        <Dropdown 
+          options={[
+            { id: "0", text: "All datasets" },
+            { id: "1", text: "Datasets where I have allocations" },
+            { id: "2", text: "My published datasets" },
```
idiom-bytes, line 302: Can we reduce the labels/text so it fits in the width of the component/element.

idiom-bytes:
LGTM!

idiom-bytes (dismissed):
Looks good to me. Thanks for cleaning up the drop down menus. Some code changes taking place in locations (i.e. purgatory) that I was not expecting.

KatunaNorbert:
@idiom-bytes Thanks for looking into this

EnzoVezzaro:
UI wise looks good, I think it's ready to be shiped. But I found some issues with my assets:

Wallet: 0xC50ce98B13Aa9782cAc4aF6809007B2b035E599e
Networl: Goerli

Screenshot 2023-02-28 at 11 17 49

With "All Datasets" i can see assets, but if I do "My published", it's empty:

Screenshot 2023-02-28 at 11 17 45

Even though I can see them on the market:

Screenshot 2023-02-28 at 11 17 30

KatunaNorbert:
@EnzoVezzaro I see that those assets from the marketplace are published on test network. In df-web we don't display assets from test networks, so probably that's why you don't see them

EnzoVezzaro:
> @EnzoVezzaro I see that those assets from the marketplace are published on test network. In df-web we don't display assets from test networks, so probably that's why you don't see them

Mmmm... I see. You're correct, they are all assets from Mumbai or Goerli. I'm confused then. Here I have all my assets (the ones that I published through the market):

Screenshot 2023-02-28 at 13 00 00
(they're many)

but if I select "My Publshed" I see empty. I should see all my published assets, correct? or just the ones that I have allocated on (in this case just one asset, the one in purgatory)? But yes, they're all on a testing network.

src/utils/data.js
```diff
@@ -14,6 +20,12 @@ export let filterDataByUserAllocation = (datasets, allocations) => {
     return filteredDatasets
 }
 
+export let filterDataByOwner = (datasets, ownerAddress) => {
+    let filteredDatasets = []
+    filteredDatasets = datasets.filter((d) => d.owner === ownerAddress)
```
EnzoVezzaro: Found the issue. Basically, our filter is not case sensitive, so it leaves out any result that doesn't match exactly. While our `d.owner` is always lowercase, `ownerAddress` is not. 

I'll add a quick fix: `datasets.filter((d) => d.owner === ownerAddress.toLowerCase())`, it should be enough.

EnzoVezzaro (approved):

EnzoVezzaro:
@idiom-bytes not sure what happened with your approval. Somehow GH decided to dismissed after I pushed some changes to this PR. Please just take a quick look at it and approve it again.

src/utils/data.js
KatunaNorbert: I see, good catch. Thanks for fixing this!

KatunaNorbert:
@EnzoVezzaro the approval was dismissed because you pushed new changes and each time we push new changes, those needs to pe approved again.
I'm going to approve and merge it.

EnzoVezzaro (approved):

PR #484, opened 2023-03-02
Fixing clock that was off by 6 hours.

### LINKED ISSUE #379: Address time frame issue in df-web - DF timeline reset at midnight while countdown clock stated over 7 days
### Problem:
- DCV amounts, and countdown timer reset due to epoch being completed, but time was off.
- Timer still shows off remaining time relative to local time rather than global UTC. There should never be more than 7 days to go.
- Round number did not increment as expected (it still reads like it's round 9, but it's round 10)

### DoD:
- [ ] Countdown timer shows time left UTC-based, not local-based
- [ ] Round does not count up

idiom-bytes:
Fix #379 
- Clock was off by 6 hours. Obvious fix.

### DIFF
--- a/src/components/common/CountDown.svelte
+++ b/src/components/common/CountDown.svelte
@@ -6,7 +6,6 @@
   export let title = undefined;
 
   let countdownEnd = moment(getThursdayDate(moment().utc()))
-    .set({ hour: 6 })
     .add(moment().utcOffset(), "minutes")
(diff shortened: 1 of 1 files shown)


KatunaNorbert:
@idiom-bytes That's strange, no idea why the hour was set to 6 :)). Thanks for the fix.

PR #488, opened 2023-03-03
Expand yield metrics reported, and make them easier to compare

### LINKED ISSUE #487: Simplify yield/return per round for the user, and average last 3, 5, 10 rounds (farms page)
### Problem:
The current APY is too complicated.
Let's add a simple yield per round for the user.
We can then approximate APY by * 52. (assuming a consistent market, which doesn't exist).
Modify the tooltip to represent this.

Other considerations - Smooth out the past average
- Rather than extrapolate out, extrapolate into "stable metrics"
- Average of the last 3, 5, 10 rounds
- Figure out a good number to use

### DoD:
- [ ] Calculate average yield for reporting in farms page

idiom-bytes:
Fixes #487
### DoD:
- [x] Keep APY front and center.
- [x] Review metrics being reported
- [x] Added roundAPR, roundYield, lastRoundAPR, lastRoundYield.
- [x] Merge from main [PR updates for RouindYield](https://github.com/oceanprotocol/df-sql/pull/46) and prepare this PR

### Additionally:
- Added In other PRs: roundVolume, lastRoundVolume
- ImproveD metrics reported so they are easier to compare...This is a first pass, there is more work to be done. [The APY/DCV/Allocation](https://github.com/oceanprotocol/df-web/issues/532) tabs should help improve this further.


KatunaNorbert:
The metrics are looking good, we should decide which one to show. Looks like `roundYield` is the most realistic metric which we can present  and can be most helpful for the user.

src/stores/data.js
```diff
@@ -12,8 +12,12 @@ export const columnsData = [
...
+  { key: "roundAPR", value: "RoundAPR", display: (roundAPR) => parseFloat(roundAPR ? roundAPR * 100 : 0).toFixed(2) + '%', tooltip: descriptions.default.tooltip_datafarming_current_round_asset_APY },
+  { key: "roundYield", value: "RoundYield", display: (roundYield) => parseFloat(roundYield ? roundYield * 100 : 0).toFixed(2) + '%', tooltip: descriptions.default.tooltip_datafarming_current_round_asset_yield },
+  { key: "lastRoundAPY", value: "LastRoundAPY", display: (lastRoundAPY) => parseFloat(lastRoundAPY ? lastRoundAPY * 100 : 0).toFixed(2) + '%', tooltip: descriptions.default.tooltip_datafarming_last_round_asset_APY},
+  { key: "lastRoundAPR", value: "LastRoundAPR", display: (lastRoundAPR) => parseFloat(lastRoundAPR ? lastRoundAPR * 100 : 0).toFixed(2) + '%', tooltip: descriptions.default.tooltip_datafarming_last_round_asset_APY},
+  { key: "lastRoundYield", value: "LastRoundYield", display: (lastRoundYield) => parseFloat(lastRoundYield ? lastRoundYield * 100 : 0).toFixed(2) + '%', tooltip: descriptions.default.tooltip_datafarming_last_round_asset_yield},
```
KatunaNorbert, line 20: We should define this new tooltip descriptions

EnzoVezzaro:
when you toggle the columns, it goes from 0.00% to 0 and then to `undefined`.


https://user-images.githubusercontent.com/13741335/222714465-69dd56b5-f4fb-4e72-8133-f45e6662d543.mp4

EnzoVezzaro:
These metrics are fine, but we keep adding information to the user. We need to filter out what's most relevant here. 

While doing some discovery I came across some products that are doing a great good displaying this information, for what I can see most platform play around with APR and APY this way:

Harvest:
Screenshot 2023-03-03 at 08 09 17
Screenshot 2023-03-03 at 08 09 39

Instadapp Lite:
Screenshot 2023-03-03 at 08 06 03
Screenshot 2023-03-03 at 08 06 29

Curve:
Screenshot 2023-03-03 at 08 14 12
Screenshot 2023-03-03 at 08 14 36

Pancake Swap:
Screenshot 2023-03-03 at 07 59 03
Screenshot 2023-03-03 at 08 04 32

Yield Finance:
Screenshot 2023-03-03 at 08 01 21

This might not be exactly our use case, but it can help us identifying what're the most important metrics to show. for what i see, simplicity is key.

jamiehewitt15:
I thought we had agreed in the meeting to get rid of the annualised numbers (APY and APR)? By taking the yield round and multiplying it by 52, we are not creating helpful information for the user - we are just extrapolating to present unreliable figures to them.

idiom-bytes:
As discussed, we still want to lead with APY.
- Room to improve visualization around variable rate. I.E. tooltips, calling it vAPY
- We need to update docs.
- We need to improve UX (i.e. add accordion, to provide details)
- Room to improve via new calcs in df-sql: avg_apy_last_3_weeks

Will move to draft until it's further along.

jamiehewitt15:
> * Room to improve visualization around variable rate. I.E. tooltips, calling it vAPY (_it's out there_)

Yeah, I'd still prefer for us to avoid APY altogether, but if we must show it then it makes sense to acknowledge the variability. 

> * Room to improve via new calcs in df-sql: avg_apy_last_3_weeks

I find this confusing. APY is an annualised rate, so why would a user want to see it averaged over a three-week period?

KatunaNorbert:
I propose the following steps:
- Keep only one metric visible by default, that could be APY, and then as mentioned above improve it by name, tooltip message etc.
- Make rows expandable and add more metrics + explanation there. We can do this following the docs: https://carbon-components-svelte.onrender.com/components/DataTable#expandable-rows.

EnzoVezzaro:
> * Make rows expandable and add more metrics + explanation there. We can do it following this docs: https://carbon-components-svelte.onrender.com/components/DataTable#expandable-rows.

Yes, this is the intention. we are working on a quick mockup for it on miro.

KatunaNorbert:
Added extendable row containing current and last round APYs

jamiehewitt15:
The arrow on the left is so small I can imagine a lot of people wouldn't notice it. I suggest making it bigger, we could also make it so that clicking anywhere in the white space also expands the row. 

image

src/stores/data.js
```diff
@@ -63,7 +67,11 @@ function getRow(dataInfo, key) {
     symbol: dataInfo.symbol,
     owner: dataInfo.owner_addr,
     lastRoundAPY: dataInfo.lastRoundAPY,
-    roundapy: dataInfo.apy,
+    lastRoundAPR: dataInfo.lastRoundAPR,
+    lastRoundYield: dataInfo.lastRoundAPR / 52,
+    roundAPY: dataInfo.apy,
+    roundAPR: dataInfo.apr,
+    roundYield: dataInfo.apr / 52,
```
KatunaNorbert, line 74: I don't think this is right. This APR is already the earning for the current round, I don't think we should divide it by 52 since it doesn't do the calculation for 1 year.

EnzoVezzaro:
Sorry for letting you know just know 😅 I just realised that your design might conflict a bit with what we're been preparing in miro. I'll post here the high level mockup we have on the board.

Screenshot 2023-03-24 at 10 07 53

KatunaNorbert:
@EnzoVezzaro Thanks for letting me know, although I don't think we want to get all those changes in this PR. Let's just add the APY/APR/Yield and keep updating the extended row in other PRs.

EnzoVezzaro:
I don't see the extended view anymore 😶

Screenshot 2023-03-28 at 10 51 38

EnzoVezzaro:
> That was removed on @idiom-bytes request

I see 👍🏽

EnzoVezzaro (dismissed):
LGTM 👍🏽

jamiehewitt15 (approved):
Looks good to me

PR #491, opened 2023-03-06
Fix allow user to update only the lock end time

### LINKED ISSUE #485: Can't update lock end time only
When selecting a new date and clicking on the button to update lock end date, an error message is displayed under lock amount saying that the lock amount should be grater than 1. It should not be connected to amount field.

KatunaNorbert:
Fixes #485.


jamiehewitt15 (approved):
Looks good to me

PR #493, opened 2023-03-06
highlight if an asset get's 2x reward (asset published by the user)

EnzoVezzaro:
- Fixes #493

Changes proposed in this PR:

- highlight if an asset get's 2x reward (asset published by the user)

Screenshot 2023-03-06 at 12 21 59

### DIFF
--- a/src/stores/data.js
+++ b/src/stores/data.js
@@ -27,7 +27,6 @@ export const columnsData = [
     display: (volume) => '$' + volume,
     tooltip: descriptions.default.tooltip_datafarming_last_round_consume
   },
-  { key: "doubleapy", value: "DoubleAPY", display: (doubleapy) => parseFloat(doubleapy ? doubleapy : 0).toFixed(2) + '%', tooltip: descriptions.default.tooltip_datafarming_double_APY },
   { key: "nftaddress", value: "NFTAddress" },
   { key: "did", value: "DID" },
   { key: "roundallocation", value:"RoundAllocation", display: (allocated) => allocated + ' veOCEAN', tooltip: descriptions.default.tooltip_datafarming_round_allocation},
@@ -77,7 +76,6 @@ function getRow(dataInfo, key) {
     symbol: dataInfo.symbol,
     owner: dataInfo.owner_addr,
     lastroundapy: dataInfo.lastRoundAPY,
-    doubleapy: convertWPRtoAPY(dataInfo.apr*2),
     roundapy: dataInfo.apy,
     nftaddress: dataInfo.nft_addr,
     ispurgatory: dataInfo.is_purgatory,


KatunaNorbert:
I'm getting an error on Farm page
Screenshot 2023-03-07 at 11 11 04
.com/37236152/223376667-3f0b0d1f-d9f0-4b47-a759-ab528eefc1d2.png">

KatunaNorbert:
I mentioned in issue that now df-sql serves boosted publisher allocation for an asset. We should also display that somewhere, I think it's a high impact for publisher.

jamiehewitt15:
Yeah, I'm getting the same error. Also, why is there a react error when we are not using React? 🤔

image

KatunaNorbert:
Ok, the error persists on data received from test env, so I inspected and discovered that is caused by some datasets having the owner as `undefined` @trizin.
I handled the issue in this PR #494.

KatunaNorbert:
Sure @trizin 
`body: JSON.stringify({
        "query":{
          round:roundNumber
        },
        "sort":{
          "volume":-1
        }
      })`

KatunaNorbert:
Looks like Aquarius has not indexed those assets yet, maybe that's why we don't get publisher related data:
https://market.oceanprotocol.com/asset/did:op:007e31cfb62eb7aed07330574fdb8dca93b892a14bd7842cd0debad016c743ee

KatunaNorbert:
@EnzoVezzaro  I already merged a fix, you can get changes from main and maybe we can iterate here for a better solution

EnzoVezzaro:
> @EnzoVezzaro I already merged a fix, you can get changes from main and maybe we can iterate here for a better solution

I already did 👍🏽 it should be already merged into this PR.

EnzoVezzaro:
wait, I see what you mean. After checking your commits, it should just be:

Screenshot 2023-03-07 at 09 13 58

but in local, I'm getting this result with your code (this is not correct, right):

Screenshot 2023-03-07 at 09 13 51

I see that in [DF test](https://test-df.oceandao.org/datafarming) owners are removed in this case.

Screenshot 2023-03-07 at 09 18 33

I'll push your fix to see what happened, it might be just something on my machine.

jamiehewitt15:
@EnzoVezzaro can you increase the z-index on the tooltip so that we can see it above the table? Overall seems to be working well now.

image

KatunaNorbert:
> I mentioned in issue that now df-sql serves boosted publisher allocation for an asset(`ve_allocated_owner, ve_allocated_realtime_owner`). We should also display that somewhere, I think it's a high impact for publisher.

@EnzoVezzaro can you also handle this here?

EnzoVezzaro:
> 

sure, I'll look into it 👍🏽

EnzoVezzaro:
@KatunaNorbert should we display these values on the table, next to "CurrentAllocation". I have this entry as example:

`
{
    "chainID": 80001,
    "nft_addr": "0x4d6bee7f7833c6e9927e72a8eeb3ade8aca8eade",
    "did": "did:op:4f8e2a408c136ae9a50768728df8b9371a97d12b32a285ae6ec03f8c920f5944",
    "symbol": "OCEAN-NFT",
    "name": "Estadisticas de Pruebas Nacionales, 2016-2021 ",
    "ve_allocated": 36.7882080078,
    "ocean_allocated": 20,
    "ve_allocated_owner": 36.7882080078, // WE NEED TO ADD THIS
    "ocean_allocated_owner": 20,
    "ve_allocated_realtime": 39.1363868713,
    "ocean_allocated_realtime": 20,
    "ve_allocated_realtime_owner": 39.1363868713, // WE NEED TO ADD THIS
    "ocean_allocated_realtime_owner": 20,
    "volume": 0,
    "is_purgatory": 1,
    "apr": 0,
    "apy": 0,
    "owner_addr": "0xc50ce98b13aa9782cac4af6809007b2b035e599e",
    "round": 0
  }
`

but i have a question first. Why is my `ve_allocated_owner` amount slightly different than `ve_allocated_realtime_owner`? what's the diffenrence between those entries?

KatunaNorbert:
@EnzoVezzaro We apply a weight function on the allocation so it increases during the week, If you allocate at the beginning of the week you will have like `1/50 * allocationPower` of allocation power visible on first day and `50/50 * allocationPower` on last day. This is so the Whales doesn't come on last second, allocate and then they have 100% allocation power. If they allocate in last second it will actually be like 1/50, If they allocate at the middle of week will be 25/50 etc.
But this is confusing for most of users, so we also have realtime allocation which show entire allocation power without applying the weighted function.

EnzoVezzaro:
> > I mentioned in issue that now df-sql serves boosted publisher allocation for an asset(`ve_allocated_owner, ve_allocated_realtime_owner`). We should also display that somewhere, I think it's a high impact for publisher.
> 
> @EnzoVezzaro can you also handle this here?

I have integrated `ve_allocated_realtime_owner` in the datatable 👍🏽 but I'm not sure it's updating correctly. Please take a look carefully.

idiom-bytes:
It is updating correctly.

This column can't be enabled by default.
"My Allocation" becomes hidden and should generally be visible.

KatunaNorbert:
I agree with @idiom-bytes, I like the highlight of the column and the tooltip but it's not to visually effective.
I would try to combine the owner allocation with the MyAllocation column, something like this
Screenshot 2023-03-10 at 14 44 08

EnzoVezzaro:
> beside the tooltip there is already a new row color for your published assets. We can maybe user a more shinny color, like you suggested a gold or something.

Great, let's do it.

> I would keep the tooltip and message, it's going to make things more clear for the user. I suggest to have the 2x allocations/rewards visible somewhere in the row.

Great suggestions @KatunaNorbert . I'll implement them, not sure yet how to display "2x allocation/reward" but a "2x" badge can work. Let's see how i'll come out.

> Ideally we also show the user allocation in veOCEAN not just percentage, I would use the same column for this to win some more space inside the table.

👍🏽

EnzoVezzaro:
CURRENT STATE:

- [x] add row state with backgrounds as mockup
- [x] add filtering by 2xers
- [x] change owner address to "owned by you" (if owner of dataset)
- [ ] add color indicator in dropdown

Fixes / Testing:

- [ ] fix jump on first input
- [ ] test, check results

KatunaNorbert:
All this different highlights for rows might be little bit too much on my opinion, and also confusing because it doesn't explain what that means until you open the dropdown, action that it's not intuitive.

I also think there is no difference between "2xers" and "My published". All your published datasets are "2xers", but of course it will only have effect if you allocate and get consume volume. We can reduce the complexity of another filter and row color or whatever else we use to just 1 instead of 2.

I would suggest that if we want to use a background highlight, then stick with only one(published assets) and also have that explained either with a label on the row or with a dot + text explanation on the top of the table.

I'm also not sure If we aiming for adding the publisher allocation in here or not.

EnzoVezzaro:
Thank you for the review @KatunaNorbert 🙏🏾 we're not yet done with it, but it's good you looked into it. 

> All this different highlights for rows might be little bit too much on my opinion, and also confusing because it doesn't explain what that means until you open the dropdown, action that it's not intuitive.

We though about it, but we considered that having repetitive elements like "icons" or "symbols" would crowd too much the design since we're trying to visualize 3 different states (allocated, rewarded (2xers), published).

> I also think there is no difference between "2xers" and "My published". All your published datasets are "2xers", but of course it will only have effect if you allocate and get consume volume. We can reduce the complexity of another filter and row color or whatever else we use to just 1 instead of 2.

We introduced a third element (2xers) to give users visibility on datasets earning active rewards. 

> I'm also not sure If we aiming for adding the publisher allocation in here or not.

We're still working on adding a "MyVeOcean" column as mockup in [MIRO](https://miro.com/welcomeonboard/eVZYVk9qYVZ0NGV6RmtrNnkxVlJvUWVkRW1tRHl6c2JGV0E1UEIxQWR1aDYxdUFsdW9hNk51T2FLeFpudWhkSnwzNDU4NzY0NTQ0NzkyMTEyODQzfDI=?share_link_id=554632904763).

EnzoVezzaro:
Maybe you can help with something @KatunaNorbert since you have more experience than me on svelte. I'm struggling with this issue. I haven't been able to fix it yet, and it's driving me kind of nuts 😅

Basically: when I use any of the inputs the first time, it does a weird jump and goes out of focus. 


https://user-images.githubusercontent.com/13741335/226610075-285dec77-28c8-4437-8b52-99f5a4415ed2.mp4

the issue seems to be [here](https://github.com/oceanprotocol/df-web/blob/feature/issue-481-highlight-2x-allocations-publishers/src/components/common/ShareInput.svelte#L24). Do you have any suggestion?

idiom-bytes:
3 pieces of feedback:
1. Fix the tooltip hiding
2. Add a tooltip to "Title" column explaining the highlight "yellow highlight == publisher 2x"
3. Remove the bolding from the title

KatunaNorbert:
@EnzoVezzaro Regarding the issue with the allocation input, when you don't have allocations on that dataset and you change the input so you have, you set the id to the dataset which causes the table to rerender, that's why the table table it's scrolling back to the initial position.

EnzoVezzaro:
Yes, I was trying to follow what this [issue](https://github.com/carbon-design-system/carbon-components-svelte/issues/1466) recommended, as DataTable in svelte has no slot or any way to pass a class to the row. I think a got idea though, let's see if works.

jamiehewitt15:
Is PR is against the staging branch, I thought we agreed to merge PRs into main?

EnzoVezzaro:
> Is PR is against the staging branch, I thought we agreed to merge PRs into main?

you're right. Probably forgot to change it after our conversation. I'll do that now.

jamiehewitt15:
@EnzoVezzaro what happened to the tooltip icon? Looks like it's changed colour so it can't be seen

image

jamiehewitt15:
Also, I remember we raised this point in the meeting but not sure if it is addressed - where is it explained what a 2xer is? I feel like that's going to confuse people

EnzoVezzaro:
> @EnzoVezzaro what happened to the tooltip icon? Looks like it's changed colour so it can't be seen
> 
> image

hi @jamiehewitt15 both points are being worked out. Just push some commits for it

jamiehewitt15:
> has changed to "2x publishers"

I'm still a bit confused about this, is there some sort of minimum threshold to get this label as a 2x publisher? Or do they get this label and the orange colour if they allocate 1 ocean to their asset?

EnzoVezzaro:
> I'm still a bit confused about this, is there some sort of minimum threshold to get this label as a 2x publisher? Or do they get this label and the orange colour if they allocate 1 ocean to their asset?

Our goal was to highlight publishers how have allocated into their own assets and are earning 2x publisher reward. You get that state (2x publisher) on any allocation by the owner of the dataset.

jamiehewitt15:
>  Display double APY if the connect walled is owner (current round & previous rounds => min(125.00, apy * 2)

Probably best if we work through an example on this but it's not always as simple as just *2 because we have a fixed maximum reward that is dispersed for each asset. We should also consider the effect on the APY for the users who are not publishers as this is a zero-sum game, while the publishers get more rewards everyone else gets less.

EnzoVezzaro:
> The publisher receives 2/3 of the rewards as their stake has increased but the non-publisher only receives 1/3. The APY goes up for the publisher but down for the non-publisher

I see. I think it best to think about it a little bit more. 

I'd move "[Display double allocation if the connected wallet is owner](https://github.com/oceanprotocol/df-web/pull/493#issuecomment-1477901714)" to a new issue and take it on another moment to ship what's been done here so far.

what do you think? @idiom-bytes @KatunaNorbert @jamiehewitt15 @trizin

src/stores/data.js
```diff
@@ -76,7 +76,7 @@ function getRow(dataInfo, key) {
     symbol: dataInfo.symbol,
     owner: dataInfo.owner_addr,
     lastroundapy: dataInfo.lastRoundAPY,
-    // doubleapy: parseFloat(dataInfo.apy * 2).toFixed(3), // TODO: enable after discussion around it
+    doubleapy: parseFloat(dataInfo.apr * 2).toFixed(3),
```
jamiehewitt15, line 79: This calculation should be double the weekly yield and then compounded across 52 weeks to get the APY

KatunaNorbert: `convertWPRtoAPY(dataInfo.apr * 2)`

KatunaNorbert:
I think we should have another background color for hovering the row because this way the input it's not really visible:
Screenshot 2023-03-27 at 15 31 02

src/stores/data.js
EnzoVezzaro: changed 👍🏽 thanks

jamiehewitt15:
How is it that the round APY is more than the Double APY for this asset at the top? Are you sure the calculations are correct?

image

src/utils/rewards.js
```diff
@@ -15,6 +15,13 @@ export const convertWPRtoAPY = (wpr) =>{
   return apy_passiv
 }
 
+export const convertWPRtoDoubleAPY = (wpr) =>{
+  const weeks = 52
+  const double_apr = wpr * 2;
+  const double_apy = ((1 + double_apr / weeks) ** 52 - 1) * 100
+  return double_apy
+}
+
```
KatunaNorbert, line 24: we already have the `convertWPRtoAPY(wpr)` doing the same thing, can we use that one? Just pass it wpr * 2 as param.
The current formula is also wrong, the double_apr should be `double_apr=wpr * 2 * weeks`

KatunaNorbert (changes requested):

KatunaNorbert:
I don't really get the purpose of doubleAPY where we are just doubling the APR. This is only going to be real If only the publisher allocates to that dataset, but also wrong in this case as @jamiehewitt15 explained above.
Only real use for doubling APY would be to do it on user level and showing userAPY vs publisherAPY.

EnzoVezzaro:
> Only real use for doubling APY would be to do it on user level and showing userAPY vs publisherAPY.

removed double apy column as discussed. This is a user level feature tackle in #539

EnzoVezzaro:
@jamiehewitt15 @KatunaNorbert can we have another review on this please

jamiehewitt15 (approved):
Looks good to me

idiom-bytes:
```removed double apy column as discussed. This is a user level feature tackle in #539````
I updated #539, the goal is not a separate column.

PR #501, opened 2023-03-09
change date format in myocean lock 

### LINKED ISSUE #496: veOCEAN date format is inconsistent: DD-MM-YYYY one place, MM-DD-YYYY another
### Description

Go to https://df.oceandao.org/veocean and you'll see something like the following. The dates are where the red arrows are. Note how they're inconsistent:
- date on the left is DD-MM-YYYY
- date on the right is MM-DD-YYYY (it's 4 years from now, in March 2027)

Did you know? One of these is European date format, the other is North American. (Yikes)

### Towards a solution

Rather than just being consistently one or the other (NA or Europe) we can do something that is both (a) consistent and (b) not vague.

Possibilities:
1. YYYY-MM-DD. E.g. 2023-03-27. Goes from big --> small, nice to sort. But, you have
2. YYYY-MonthName-DD. E.g. 2023-Mar-27
3. MonthName-DD-YYYY. E.g. Mar-27-2023
4. DD-MonthName-YYYY. E.g. 27-Mar-2023

Analysis:
- (1) leaves room for confusion (MM <> DD)
- (2) leaves no room for confusion and is a sane order (big -> med -> small)  
- (3) leaves no room for confusion and is how people often think of dates: "March 27, 2023" 
- (4) leaves no room for confusion and is a sane order (small -> med -< big)

(2)(3)(4) are basically tied for best. (1) may cause confusion.

EnzoVezzaro:
- Fixes #496 .

Changes proposed in this PR:

- change date format (month)

### DIFF
--- a/src/components/claim/EpochHistory.svelte
+++ b/src/components/claim/EpochHistory.svelte
@@ -28,7 +28,7 @@
       )
     );
     rows.forEach((row) => {
-      row.date_start = moment(row.date_start).format("DD-MM-YYYY");
+      row.date_start = moment(row.date_start).format("DD-MMM-YYYY");
       row.passive = `${row.passive} OCEAN`;
       row.active = `${row.active} OCEAN`;
     });
--- a/src/components/veocean/VeOceanCard.svelte
+++ b/src/components/veocean/VeOceanCard.svelte
@@ -69,9 +69,9 @@
         {loading}
       />
       <ItemWithLabel
-        title={`Lock ends`}
+        title={`Lock ends`} 
         value={`${
-          $oceanUnlockDate ? $oceanUnlockDate.format("DD-MM-YYYY") : "-"
+          $oceanUnlockDate ? $oceanUnlockDate.format("DD-MMM-YYYY") : "-"
         }`}
         tooltipMessage={descriptions.default.tooltip_veocean_my_lock_ends}
         {loading}


EnzoVezzaro:
not sure why I'm trailing the plausible changes to staging, do we have analytics on staging @jamiehewitt15 ?

jamiehewitt15:
> not sure why I'm trailing the plausible changes to staging, do we have analytics on staging @jamiehewitt15 ?

nope, only on the live site

jamiehewitt15:
Can you do a pull from main?

jamiehewitt15:
Why is this merging into staging? To close the issue, we need the fx to be merged into main?

EnzoVezzaro:
> Why is this merging into staging? To close the issue, we need the fix to be merged into main?

shouldn't we follow a clean path for issues? Local -> Staging -> Production

I can change branch again, but we need to agree on the process. @jamiehewitt15 @KatunaNorbert

jamiehewitt15:
> shouldn't we follow a clean path for issues? Local -> Staging -> Production
> 
> I can change branch again, but we need to agree on the process. @jamiehewitt15 @KatunaNorbert

Yeah, I don't think we're using that staging branch anymore

KatunaNorbert:
In order for the changes from main to not be visible here, we should merge main into staging.
We can keep staging for people that want to play with testing env, but update it from main after something is merged in main.

KatunaNorbert:
I updated staging, but the changes didn't go away, probably this will only happen if we close PR and open again

EnzoVezzaro:
> We can keep staging for people that want to play with testing env, but update it from main after something is merged in main

I see. Ok, then. I'll change branch to main to have it merge there.

EnzoVezzaro:
@jamiehewitt15 @KatunaNorbert 👍🏽

maybe we start thinking about having staging updated, and make main just our production branch which gets updated from staging. I know we're a small team and might not be necessary, but I think that since we're changing some internal processes, this might be beneficial. If we follow a dev > staging > production process, we can bundle more fixes and features into releases that we can announce before hand. It's just a thing to think about. 😉

KatunaNorbert:
@EnzoVezzaro If we choose to go with a new format `DD-MMM-YYYY` can we be consistent and also use this format in other places?
Does the new format solves the inconsistent format issue, or that was just a misunderstanding from the user, mixing up the month with the day?

jamiehewitt15:
Yeah, I think this is the best format. We just need to be consistent everywhere

EnzoVezzaro:
On #496 we're fixing an issue related to veOcean page. This is just a small adjustment to make this page a bit less confusing. It was supposed to be a quick fix that we could ship fast. 

Now, if we want to change the formar elsewhere, we can do it. But I have only identify this case:

Screenshot 2023-03-13 at 08 15 10

Is there any other places (beside the input, I wouldn't mix numbers/letters on that input)?

jamiehewitt15:
> Is there any other places (beside the input, I wouldn't mix numbers/letters on that input)?

I don't think so

KatunaNorbert:
@EnzoVezzaro inside #496 there is a comment about date being inconsistent DD-MM on one place and MM-DD on other, I think that was the first issue mentioned:
"date on the left is DD-MM-YYYY
date on the right is MM-DD-YYYY (it's 4 years from now, in March 2027)".

I'm wondering if that issue still persist. If does then it's fine to change only the format here but then we should have another issue for that. If we close the current issue it means that we fixed both.

KatunaNorbert:
If it's just a wrong interpretation by user because of DD-MM and it's solved with DD-MMM format, than it's fine

PR #534, opened 2023-03-22
Added lastRoundVolume

### LINKED ISSUE #533: Make sure roundVolume and lastRoundVolume are available
- [x] Double checked RoundVolume was still there
- [x] Added lastRoundVolume

idiom-bytes:
Fixes #533 
- Verified roundVolume was still available
- Added lastRoundVolume as another data point

### DIFF
--- a/src/components/common/Table.svelte
+++ b/src/components/common/Table.svelte
@@ -4,12 +4,16 @@
   import Toolbar from "carbon-components-svelte/src/DataTable/Toolbar.svelte";
   import ToolbarContent from "carbon-components-svelte/src/DataTable/ToolbarContent.svelte";
   import ToolbarSearch from "carbon-components-svelte/src/DataTable/ToolbarSearch.svelte";
-  import "carbon-components/scss/components/_data-table.scss"
+  import "carbon-components/scss/components/_data-table.scss";
   import Button from "./Button.svelte";
-  import 'carbon-'
+  import "carbon-";
   import ChecklistDropdown from "./ChecklistDropdown.svelte";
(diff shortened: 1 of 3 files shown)


EnzoVezzaro (dismissed):
LGTM 👍🏽

KatunaNorbert (dismissed):

KatunaNorbert:
Right, it's the same for me, you need to refresh otherwise it's undefined. Let me look into this.

EnzoVezzaro:
> Are you sure this is correct? Says `undefined` for every asset... If they didn't have any volume last round it should be 0 rather than `undefined`:
> 
> image

yeah, I'm having the same issue. Also when you remove and add a column the tooltip on that column disappears (this doesn't seems related to this PR, but maybe it can be worked out in here).

EnzoVezzaro (approved):
great job @KatunaNorbert 👏🏽 not issues found

jamiehewitt15 (approved):
Yeah looks good now

idiom-bytes:
```Are you sure this is correct? Says undefined for every asset... If they didn't have any volume last round it should be 0 rather than undefined:```

I worked on it after seeing the issue so by the time I ran again, it all seemed to work. I thought it had to do with fresh assets without a lastRound.value, so I bound it to a min value of zero, and it all seemed to work.

Checking with Norbert..

idiom-bytes:
I think I can just check for typeof !== undefined.
Will validate asap.

KatunaNorbert:
@idiom-bytes Yes, I fixed the issues. I'm merging this one.

PR #548, opened 2023-03-29
Simplify rewards component

### LINKED ISSUE #543: Simplify Passive + Active Rewards Component
Rewards Page

Change our current Simplify Passive + Active Rewards components for a single component for FTUE-only.

KatunaNorbert:
Fixes #543.


KatunaNorbert:
I propose merging this one as it is and postpone further development until we have more data and we can proceed with improvements.
Other option will be to just close it.

idiom-bytes:
I have made some further improvements as part of this PR.
You can view them in the image below.

Screenshot from 2023-04-04 12-30-59

idiom-bytes (approved):
I have made some adjustments and updates. LGTM! Let's ship!

src/components/claim/ClaimRewardsVeDF.svelte
```diff
@@ -75,112 +76,123 @@
...
+    <p>
+      Data Farming consists of two reward programs. 
+      Each reward program is unique and offers different ways for 
+      participants to get engaged. To access Reward Programs, 
+      participants are required to hold veOCEAN.
+    </p>
+    <div class="rewardsContainer">
+      <ClaimItem
+        title="Passive"
+        description="<p>Earn Passive Rewards from Data Farming by <strong>locking OCEAN</strong> and <strong>holding veOCEAN</strong>.</p><p><strong>Earn more Rewards</strong> by locking more OCEAN or increasing your unlock time.</p>"
```
jamiehewitt15, line 90:
```suggestion
        description="<p>Earn passive rewards from Data Farming by <strong>locking OCEAN</strong> and <strong>holding veOCEAN</strong>.</p><p><strong>Earn more rewards</strong> by locking more OCEAN or increasing your unlock time.</p>"
```

idiom-bytes: We have capitalized these as Nouns because this is what the reward programs are called.

src/components/claim/ClaimRewardsVeDF.svelte
```diff
@@ -75,112 +76,123 @@
...
+                $APYs
+                  ? $APYs?.activeUser > 10000
+                    ? "over 10000"
+                    : `${$APYs?.activeUser.toFixed(2)}`
+                  : 0
+              }% Your APY`
+            : ""
+        }`}
+        apyTooltip={descriptions.default.tooltip_rewards_apy_active}
+        description="<p>Earn Active Rewards from Data Farming by <strong>allocating veOCEAN</strong> and <strong>curating quality data</strong>.</p><p>Get <strong>2X Rewards by publishing</strong> your own datasets and allocating to them.</p>"
```
jamiehewitt15, line 161:
```suggestion
        description="<p>Earn active rewards from Data Farming by <strong>allocating veOCEAN</strong> and <strong>curating quality data</strong>.</p><p>Get <strong>2X rewards by publishing</strong> your own datasets and allocating to them.</p>"
```

idiom-bytes: We have capitalized these as Nouns because this is what the reward programs are called. I'm going to resolve this feedback as complete.

src/components/claim/RewardOverview.svelte
```diff
@@ -43,7 +42,21 @@
...
   }`}
   tooltipMessage={descriptions.default.tooltip_rewards_overview}
-  className="rewardsOverview"
 >
   <Countdown />
+  <p class="cardFooter">
+    Data Farming Rounds start and end every Thursday at midnight GMT.
+    Rewards are paid in OCEAN and distributed every week on Thursday.
+    Ocean Emissions are designed to take place over the course of decades
+    while incentivizing active protocol contributors to flourish.
```
jamiehewitt15, line 51:
```suggestion
    Data Farming rounds start and end every Thursday at midnight GMT.
    Rewards are paid in OCEAN and distributed every week on Thursday.
    Ocean emissions are designed to take place over decades
    while incentivizing active protocol contributors to flourish.
```

src/components/claim/RewardOverview.svelte
```diff
@@ -19,9 +19,8 @@
 </script>
 
 <Card
-  title={`Round ${roundInfo?.id}  -  ${
-    parseInt(roundInfo?.passive) + parseInt(roundInfo?.active)
-  } OCEAN`}
+  title={`Round ${roundInfo?.id}`}
```
jamiehewitt15, line 22: Given that this is the main heading for the whole site I think it makes sense to have "Data Farming" in the title

```suggestion
  title={`Data Farming Round ${roundInfo?.id}`}
```

KatunaNorbert:
The only thing I would change is adding back the Passive Reward card description about rewards being distributed for last week holdings. For the first week user is only going to get active rewards so that should be pretty important to explain.

idiom-bytes:
```The only thing I would change is adding back the Passive Reward card description about rewards being distributed for last week holdings. For the first week user is only going to get active rewards so that should be pretty important to explain.```

This happens for both reward types. It's not special to Passive Rewards.
It's right before a key CTA, it's not the place for it.

I'll place this at the top paragraph and adjust the copy.

idiom-bytes:
Ok, so for the sake of reducing text and improving the UX/UI, I have:
1. Updated the **[DF Round Component]** tooltip to link to the docs that relate to Rounds/Rewards every thursday.
2. [Updated the docs #1200](https://github.com/oceanprotocol/docs/pull/1201) to communicate more about the time (GMT), and details such as your carrying veOCEAN balance. < Please Review

We'll create additional tasks to improve the UX/UI to communicate decreasing veOCEAN, and many other systems. We'll continue to improve it further on another task. Let's ship this.

idiom-bytes:
Hi Norbert, I'm just reviewing the macro, planning, and changes done on here **and they do not address the mockup/design.**

The goal was to output a FTUE-friendly component which is not the case here.
Screenshot from 2023-04-06 16-25-39

I'm going to rename this ticket to what it is (restructuring the cards) which is not the priority.
We'll still get this rolled out.

KatunaNorbert:
@idiom-bytes Well this was the design when the PR was created but then was changed on the run.
To simplify it based on the new design from the photo above it's going to need some more changes, like moving claim rewards functionality into another component

EnzoVezzaro:
I'm not fan of the box shadow on the container with the rewards at all. I'd remove it completelly, and just leave the title with text with background transparent.

Screenshot 2023-04-10 at 10 33 28

Also I agree with @idiom-bytes this doesn't seem to address the main concern (simplifying rewards component). 

If this helps, we might just want to simplify the design by merging these cards, like so:

Connected:
Screenshot 2023-04-10 at 10 31 03

Not connected:
Screenshot 2023-04-10 at 10 31 40

We need also to take into consideration all the changes we're implementing on the home. To have an idea I made a quick mock showing all the changes proposed for the homepage so far. You'll find a copy on miro. 

Group 32 (1)

KatunaNorbert:
@EnzoVezzaro I'm also not a big fan of cards inside cards, this has been @idiom-bytes request.
Your mock looks pretty good, but bear in mind that If we change the card style, we should be consistent and change it everywhere. For example in the banner also have a big font size for APY, left align etc. I would also have 1 button for each card, that's needed for claiming rewards and also you can do 2 different actions after you locked, increase lock and allocate, we shouldn't encourage just 1.

There is also a separate issue now specific for **simplifying** the component, we can reconsider more things inside that issue.

jamiehewitt15:
Yeah, I think @EnzoVezzaro has made some good recommendations, maybe we can discuss them in the UX/UI meeting today?

EnzoVezzaro:
Let's just finish it up for now. We can iterate later. I'd still suggest to remove the outter box of the "reward program" component.

KatunaNorbert:
Sound good. I removed the outside box.

EnzoVezzaro (approved):
LGTM

jamiehewitt15 (approved):
Looks good to me

PR #550, opened 2023-03-29
tabs component on farm page

### LINKED ISSUE #532: FE - Build tab components for APY/DCV/Allocation & Wire all the data onto UI
Reference issue #488

## APY
- Cur APY
- Last APY
- Avg Last 3
- Avg Last 5
- Avg Last 10

## DCV
- Cur Volume
- Last Volume
- Avg Last 3
- Avg Last 5
- Avg Last 10

## Allocation
- Cur Alloc
- Last Alloc
- Avg Last 3
- Avg Last 5
- Avg Last 10

### DoD:
- [x] Let's review our documentation and make sure all metrics are explained to the user
- [x] There should be a default, FTUE state, so it's easy for users to make decisions
- [x] Implement APY/DCV/Allocation tabs
- [x] Implement new columns/data being served
- [x] Every tab should have "My Allocation" available so it's easy to farm

EnzoVezzaro:
- Fixes #532 .

Changes proposed in this PR:

- add tab component on farm page table
- remove column selection

WIP:

- [x] update tooltips
- [x] make sure "my allocation" is always visible
- [x] review columns name
- [x] test component / values
- [ ] test against [test-df-sql](https://github.com/oceanprotocol/df-sql/issues/49)

FOR TESTING:

- change `NODE_ENV` to  "production" in`.env` file (until [test-df-sql](https://github.com/oceanprotocol/df-sql/issues/49) is solved)


jamiehewitt15:
It's a little bit hard to see which tab you are on because the difference between light grey and dark grey is quite subtle. Can we use the darker grey that's being used in the bar below? 

image

jamiehewitt15:
Ro5 APY is zero for every asset which doesn't seem correct.

image

jamiehewitt15:
The allocation averages also don't seem to be correct. The current allocation is not zero but the averages are all zero:

image

image

EnzoVezzaro:
Got it 😅 sorry, test this PR in local please.

> * change `NODE_ENV` to  "production" in`.env` file (until [test-df-sql](https://github.com/oceanprotocol/df-sql/issues/49) is solved)

KatunaNorbert:
@EnzoVezzaro loading al the data takes quite long, something like 3x previous time. Do you think we can optimize it?

KatunaNorbert:
There are a few fields that were available before but now with the new structure they are not, like: nftAddress, symbol, did.

KatunaNorbert:
I think that the `Ro` abbreviation for `Round` it's not right / suggestive, I propose we either use `RND` which is the most common one or `Rd` if we want a shorter version.

KatunaNorbert:
The default view of the table is just 0 values, this could just message the user that there's nothing going on here.
Screenshot 2023-03-31 at 10 23 10


I propose that we sort the table values by default based on active tab, for example for allocations view sort desc by allocation, or by 2x, something that also shows some numbers and activity:
Screenshot 2023-03-31 at 10 23 30

EnzoVezzaro:
> I think that the `Ro` abbreviation for `Round` it's not right / suggestive, I propose we either use `RND` which is the most common one or `Rd` if we want a shorter version.

great catch 👍🏽 I was following our mock in miro.

EnzoVezzaro:
> There are a few fields that were available before but now with the new structure they are not, like: nftAddress, symbol, did.

This was discussed during our ui discussions. On the mock design (Miro) we have the tab component which reports quick relevant information, and if you want to get more in depth, you'll be able to by expanding the row to see more information (WIP) about the dataset (datanft) and publisher.

EnzoVezzaro:
> The default view of the table is just 0 values, this could just message the user that there's nothing going on here.

Are you testing on local with 'production' in .env file? If not, please do that first. In main is not really 0, but still I'll see what I can do.

EnzoVezzaro:
> I propose that we sort the table values by default based on active tab, for example for allocations view sort desc by allocation, or by 2x, something that also shows some numbers and activity:

this would be problematic for the people trying to analysing datasets. For example if I want to see the dataset "BTC/USDT" APY, and then you want to see its Allocation, it'd be pretty annoying for the user if we change the datasets when he change tabs.

EnzoVezzaro:
> > loading al the data takes quite long, something like 3x previous time. Do you think we can optimize it?
> 

UPDATES: I tried to opt as much as I could (now loads in 3-4sec compared to 2-3sec before), not sure if we can do more about this as the aggregation query actually takes some time to get back.

Screenshot 2023-03-31 at 09 18 22

KatunaNorbert:
> > > loading al the data takes quite long, something like 3x previous time. Do you think we can optimize it?
> 
> UPDATES: I tried to opt as much as I could (now loads in 3-4sec compared to 2-3sec in main), not sure if we can do more about this as the aggregation query actually takes some time to get back. Also consider that in test-df-sql might be worse due to it not being updated yet.
> 
> Screenshot 2023-03-31 at 09 18 22

Cool data it's loading like 2x faster now :D

KatunaNorbert:
> > I propose that we sort the table values by default based on active tab, for example for allocations view sort desc by allocation, or by 2x, something that also shows some numbers and activity:
> 
> this would be problematic for the people trying to analysing datasets. For example if I want to see the dataset "BTC/USDT" APY, and then you want to see its Allocation, it'd be pretty annoying for the user if we change the datasets when he change tabs.

Good point, maybe we can have then something else than DCV until we actually get some value for that, although switching to production looks like there is some allocations/volume

src/stores/data.js
```diff
@@ -3,47 +3,170 @@ import { getNetworkDataById, userAddress } from "./web3";
...
+}
+
+export const defaultColumns = {
+  'alloc': ["Title", "Rnd5 Avg Alloc", "Rnd3 Avg Alloc", "Rnd Alloc", "Curr Alloc", "My Allocation"],
+  'dcv': ["Title", "Rnd10 DCV", "Rnd5 DCV", "Rnd3 DCV", "Last Rnd Volume", "Rnd Volume", "My Allocation"],
+  'apy': ["Title", "Rnd5 APY", "Rnd3 APY", "Last Rnd APY", "Rnd APY", "Rnd Yield", "My Allocation"],
+}
 
-export const defaultColumns = ["Title", "RoundVolume", "RoundAPY", "LastRoundAPY", "CurrentAllocation", "MyAllocation", "OwnerAllocation", "MyVeOcean"]
+async function getDatasetsAvgs(api,roundNumber) {
```
KatunaNorbert, line 61: Looks like there are a few issues with this request:
- the returned data is too big and can't be received
- the average doesn't seem to be right

EnzoVezzaro:
@KatunaNorbert is there anything left here to do?

jamiehewitt15 (approved):
Looks good to me. Adds everything from the ticket and it all seems to work well

KatunaNorbert:
I don't think so, avg should be fine although on multiple datasets I see the same value trough averages

EnzoVezzaro:
> although on multiple datasets I see the same value trough averages

what do you mean? It's hard to tell, I know 😅 

but I checked, the call is correct, and the data we're showing it's the actual data we're receiving. At least that's what I can see. 

I'll double check, but @jamiehewitt15 do you see anything strange with the data, can you please check also on df-sql.

jamiehewitt15:
Interesting, I'll check the queries

jamiehewitt15:
So looks like the issue here is with the queries on the df-sql side, it's not calculating the averages for 3, 5 & 10 rounds. What you're seeing there is the average for all of the rounds that we have requested (10). The only solution I can see right now would be to return the values for each of those rounds and then we calculate the averages on the front end

EnzoVezzaro:
we had refactored the queries to get the averages calculated correctly.

KatunaNorbert:
Screenshot 2023-04-12 at 18 02 30

This avg APY for last 5 rounds looks quite high considering that in the last 3 rounds the APY was 0

EnzoVezzaro:
> Screenshot 2023-04-12 at 18 02 30
> 
> This avg APY for last 5 rounds looks quite high considering that in the last 3 rounds the APY was 0

yeah, sorry for making you retest it. We notice some weird numbers too, we're investigating

EnzoVezzaro:
Hi guys. I couldn't find a solution for the grouping issue. @jamiehewitt15 did you find a way to fix it?

jamiehewitt15:
> Hi guys. I couldn't find a solution for the grouping issue. @jamiehewitt15 did you find a way to fix it?

It's possible in multiple queries but I haven't seen a way yet to combine everything

jamiehewitt15:
So the options that I can confirm will work are: 
1. Request all of the data for every round and then calculate the averages on the front end. That requires quite a bit of computation so it will slow things down. 
2. Request the all data for the current round (as you are doing currently) and then send a separate request for the averages.

EnzoVezzaro:
> So the options that I can confirm will work are:
> 
> 1. Request all of the data for every round and then calculate the averages on the front end. That requires quite a bit of computation so it will slow things down.
> 2. Request the all data for the current round (as you are doing currently) and then send a separate request for the averages. And then combine the data

I see. Both solutions seems to me that will slow down the rendering of the component. I'll see what I can do to make viable.

KatunaNorbert:
I think I managed to fix it, let me push my changes

KatunaNorbert:
I also added an extra function to get the values for a specific nft for last 5 round so I could manually test the avg result.
Seems to be fine, you can use the function which logs the results to test and remove it after.

src/stores/data.js
```diff
@@ -93,12 +93,32 @@ async function getDatasetsAvgs3Rounds(api,roundNumber) {
...
+      headers: {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        "query":{
+          "round": {
+            "$in": getRoundsDatafarm(roundNumber, 5)
+          },
+          "nft_addr":"0x9723488dc1524849a82917a61a38bbe24a8219c1"
```
jamiehewitt15, line 121: Why are you requesting for this specific address?

KatunaNorbert: For testing purpose

EnzoVezzaro:
Great @KatunaNorbert 🚀 it looks like the sorting  ( `"sort": {"volume": -1 }` ) was causing the duplicated records. Can you guys just recheck we're getting the avgs agregated. It looks to me it's fine, I'm only seen one record x nft_addr.

KatunaNorbert:
Yeah I think both grouping and sorting. It was looking good for me, I could recheck if you did some changes to query.

src/app.config.js
```diff
@@ -27,7 +27,7 @@ export const setupAppConfig = () => {
         case "staging":
             console.log("NODE_ENV has been set to staging")
             process.env.SUPPORTED_CHAIN_IDS = "[5]"
-            process.env.BACKEND_API = "https://test-df-sql.oceandao.org"
+            process.env.BACKEND_API = "https://df-sql.oceandao.org"
```
KatunaNorbert, line 30: Looks like we can merge this PR after changing back the staging url

KatunaNorbert (changes requested):

KatunaNorbert (approved):

jamiehewitt15:
> Can you guys just recheck we're getting the avgs agregated. 

It's quite hard to confirm when so many of the values are just 0

KatunaNorbert:
@jamiehewitt15 we just switch back the staging to point to staging, where there are no values. You can change locally to point to staging if you want to test the production data.

jamiehewitt15:
Correct me if I'm wrong but this doesn't look right. The current allocation is >0 but the averages are all zero 🤔

image

EnzoVezzaro:
> Correct me if I'm wrong but this doesn't look right. The current allocation is >0 but the averages are all zero 🤔
> 
> image

as @KatunaNorbert mentioned "we just switch back the staging to point to staging, where there are no values. You can change locally to point to staging if you want to test the production data."

you need to test this locally changing the env to production.

jamiehewitt15 (approved):
Ok yeah, I've tested this locally and it looks good. Let's get it merged 👍

PR #553, opened 2023-04-04
posthog integration

### LINKED ISSUE #540: Setup posthog
Switch from plausable to posthog. 

context:
- we already had integrated plausable into df-web [#486](https://github.com/oceanprotocol/df-web/issues/486)
- ocean site already uses posthog

jamiehewitt15:
Fixes #540 

Changes proposed in this PR:

- Setting up posthog
- Removing plausible, including class names


idiom-bytes (approved):
Looks great, I'm glad we were able to cleanup all of those plausible hooks and can just use the framework.

PR #563, opened 2023-04-12
Display round distributed rewards amount

### LINKED ISSUE #552: Display round distributed rewards amount
Inside round history table display the distributed passive and active rewards

- [x] calculate and add the distributed active and passive rewards amount inside the current Passive Rewards and Active Rewards columns
- [x] add tooltips to explain both values displayed inside that column
- [x] submit PR for review
- [x] reviewed
- [ ] ship y/n? Roberto + Norbert decide offline together, one way or the other, and then execute

KatunaNorbert:
Fixes #552.

### DIFF
--- a/src/app.config.js
+++ b/src/app.config.js
@@ -27,7 +27,7 @@ export const setupAppConfig = () => {
         case "staging":
             console.log("NODE_ENV has been set to staging")
             process.env.SUPPORTED_CHAIN_IDS = "[5]"
-            process.env.BACKEND_API = "https://test-df-sql.oceandao.org"
+            process.env.BACKEND_API = "https://df-sql.oceandao.org"
             process.env.SUBGRAPH_API = "https://v4.subgraph.goerli.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph"
             process.env.VE_SUPPORTED_CHAINID = 5
             process.env.DEBUGGING = "enabled"
--- a/src/components/claim/EpochHistory.svelte
+++ b/src/components/claim/EpochHistory.svelte
@@ -156,7 +156,7 @@
   }
 </script>
 
-<h2 class="title">Data Farming History</h2>
+<h2 class="dfHistoryTableTitle">Data Farming History</h2>
 <div class="epochHistoryTableContainer">
   <DataTable sortable {headers} {rows} class="customTable">
     <svelte:fragment slot="cell-header" let:header>
@@ -203,7 +203,7 @@
   .cellContainer {
     height: auto;
   }
-  .title {
+  .dfHistoryTableTitle {
     font-weight: bold;
     width: 100%;
     margin-bottom: calc(var(--spacer) / 2);
--- a/src/components/claim/EpochHistory.svelte
+++ b/src/components/claim/EpochHistory.svelte
@@ -30,13 +30,21 @@
       value: "Round",
     },
     { key: "date_start", value: "Start Date" },
-    { key: "passive", value: "Passive Rewards" },
+    {
+      key: "passive",
+      value: "Passive Rewards",
+      tooltip: descriptions.default.tooltip_rewards_passive_history,
+    },
     {
       key: "passiveapy",
       value: "Passive APY",
       tooltip: descriptions.default.tooltip_rewards_apy_passive_history,
     },
-    { key: "active", value: "Active Rewards" },
+    {
+      key: "active",
+      value: "Active Rewards",
+      tooltip: descriptions.default.tooltip_rewards_active_history,
+    },
     {
       key: "activeapy",
       value: "Active APY",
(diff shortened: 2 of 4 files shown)


jamiehewitt15:
I find the way the information in the table is presented a bit confusing and it requires the user to read every single tooltip to figure out the information. The active and passive reward columns have the format (distributed / available) while the passive and active APY columns have the format (average/personal). I'd suggest that some of this information needs to be removed or moved to its own column to make it easier for people to digest. 

Possible ways to changes to simplify it: 
- Create separate columns for "average passive APY" and "your passive APY"
- Create separate columns for "average active APY" and "your active APY"
- The passive rewards could just show the distributed amount as this is pretty much exactly the same as the available amount in every case so showing both doesn't add much.
- If you need more space in the table, you don't need to say OCEAN after every amount, you could add a note at the top or bottom saying "Distributed amounts are displayed in OCEAN"

image

KatunaNorbert:
@jamiehewitt15 Good feedback.
This is my view:
- Keep APY columns as it is, this also aligns with other places where we display APYs, we have AVG / USER so we can keep as it is to make it consistent. Adding more columns just won't be visible because they are not going to fit inside the table.
- If we display distributed active rewards we should also display distributed passive even if it's always mostly equal to the available amount. If we only display distributed active and it's a low value, users might wonder how much passive it's then distributed, so they can now be sure that all passive rewards are distributed. 
- We can differentiate the Rewards from APYs in some way, maybe add a `|` sign instead of `/`, or regroup the columns into available rewards and distributed rewards.
- I wouldn't remove the OCEAN at the end of the values from the row. Wherever we have numbers inside tables in app we have also what it represent along side. I think this is a good practice and not having it in some places might just create confusion.

jamiehewitt15:
> or regroup the columns into available rewards and distributed rewards.

Yeah I think I would go with this option, so there could be one additional column for available rewards e.g:

| Other Column      | Available Rewards |
| ----------- | ----------- |
| ...      | 75000 \| 75000 OCEAN       |
| ..   | 65000 \| 65000 OCEAN          |
| ..   | 55000 \| 55000 OCEAN          |

jamiehewitt15:
> Keep APY columns as it is, this also aligns with other places where we display APYs, we have AVG / USER so we can keep as it is to make it consistent. Adding more columns just won't be visible because they are not going to fit inside the table.

This doesn't seem correct, in other places we have used `|` which I'd say is a better and clearer differentiator. I'd suggest we use `|` in the APY column and then split up the available and distributed rewards as suggested above. 

image

KatunaNorbert:
@jamiehewitt15 does this look better?

Screenshot 2023-04-14 at 13 07 23

jamiehewitt15:
Yeah, I think that's better. Do you think there would be space to split up the active/passive rewards distributed into two separate columns? It would make it even easier to read if they were seperate

KatunaNorbert:
We could display them, but then the table looks crowded:

Screenshot 2023-04-14 at 13 52 36

jamiehewitt15 (approved):
Looks good to me

idiom-bytes:
I believe the landing page should be succinct and attempt to convert the users.

This update adds a lot of complexity without much bang. If the APY is 0%, then obviously not a lot of rewards were dispensed. These insights are derivative and therefore redundant.

I really don't think we should submit this:
- Adds redundant information
- Adds many columns/data/info to the landing page to increase complexity

KatunaNorbert:
I agree with the APY and rewards distributed being redundant If there is 0 consume volume, but we shouldn't only relay on this case. Let me give some points:
- the APY we are showing is **avg APY** if wallet it's not connected, which it's very low value if consume it's low, but there can still be some rewards to win If you stake to the right asset/assets, which right now you can see only If you have done that for pervious round using the **user APY** value, but others won't see the potential. So basically If you are not doing a good job on allocating you still want see the max return you can get and fight for.
- if consume volume kicks in, users can see how that affects the rewards that are being distributed. If amount is low then even If I do my best with allocating, there is no chance to get some good rewards. If rewards are good, I can try to fight for bigger share of those rewards. If all the rewards are distributed then I know that now only the way users manage allocations can make the difference + the publisher rewards part.
- looking at the passive side, this can confirm that we are distributing all the passive rewards. Just give them more confidence with understandable data.

idiom-bytes:
```
the APY we are showing is avg APY if wallet it's not connected, which it's very low value if consume it's low, but there can still be some rewards to win If you stake to the right asset/assets, which right now you can see only If you have done that for pervious round using the user APY value, but others won't see the potential. So basically If you are not doing a good job on allocating you still want see the max return you can get and fight for.
```
I would agree with either showing the "Amount Available" or the "Amount Dispensed".

Not both on separate columns.
I also believe it looks complex for both to be on the same column.
TLDR; That area is getting busy and it should not be.

```
if consume volume kicks in, users can see how that affects the rewards that are being distributed. If amount is low then even If I do my best with allocating, there is no chance to get some good rewards. If rewards are good, I can try to fight for bigger share of those rewards. If all the rewards are distributed then I know that now only the way users manage allocations can make the difference + the publisher rewards part.
```

Same TLDR; This type of information is too granular for a table on the landing page. 

We already share how many rewards will be distributed and the APY, so users know if they should "fight for a bigger share". 

Example, we just kicked up the rewards and passive APY went up by 2x!
That's great, and already there.

When it comes to understanding active rewards, I believe the graph on the farming page is how to resolve these insights you're looking for. Since we're not doing them right now, I would not put this on the landing page.

PR #571, opened 2023-04-24
Improve approval awareness

### LINKED ISSUE #570: Approve token messaging and warnings
- [x] Add a warning message before approval tx that informs user about approving more than he is going to lock.
- [x] If user has approved amount show a warning to lock or click on a button to set the allowance to 0.
- [ ] Review done

KatunaNorbert:
Fixes #570.

### DIFF
--- a/public/global.css
+++ b/public/global.css
@@ -189,3 +189,15 @@ body button:focus {
 .bx--pagination {
   background-color: var(--brand-white) !important;
 }
+
+.bx--toast-notification{
+  position: absolute;
+  z-index: 1001;
+  top: 0;
+  left: 0;
+  width: 50vw !important;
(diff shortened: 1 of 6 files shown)


KatunaNorbert:
I'm just thinking If might be better to show something like >1000 for the accounts that got 'infinite' approval, instead of a very big number
Screenshot 2023-04-26 at 09 13 43

KatunaNorbert:
This is what i have done initially, but then I thought it's going to give more context to the user.
For example if I have 10 tokens left that are approved It might now worth paying the tx fee, or maybe I rather choose to lock them.

KatunaNorbert:
What about 1,000,000?

PR #581, opened 2023-06-05
Implement rewards substreams

### LINKED ISSUE #580: Implement new rewards page design including substreams
### DoD:
- [x] Design is completed
- [x] UI page is updated. Total rewards are being summed and claim buttons are adjusted.
- [x] Rewards (config file) and other details need to be updated
- [x] Resolve payment amounts and configuration
- [x] Dataflows need to be updated to get rewards from df-sql.
- [ ] Clean up any remaining issues to complete the task.

KatunaNorbert:
Fixes #580.


idiom-bytes:
Please, do not label this VolumeDF.

VolumeDF is an internal name... Let's please come up with a friendly, user-facing term.

src/App.svelte
```diff
@@ -229,10 +229,10 @@
   <WalletConnectModal />
   <main>
     <Header />
-    <Route path="/activerewards" primary={false}>
+    <Route path="/rewards" primary={false}>
       <ClaimPortal />
     </Route>
-    <Route path="/datafarming" primary={false}>
+    <Route path="/volumedf" primary={false}>
```
idiom-bytes, line 235: 1. I think we should do these as part of separate update.
Changing routes can break all sorts of things, so we need to do a seprate pass and verify everything is alright.

2. I don't recommend us changing from datafarming => volumedf... Volume df is an internal name. As documented in the ticket.

src/utils/metadata/descriptions.json
```diff
@@ -1,45 +1,45 @@
...
+  "tooltip_rewards_apy_passive": "Data Farming Passive Rewards APY is calculated relative to your Ocean locked.\n\nThe longer you lock, the higher your APY will be.",
+  "tooltip_rewards_apy_active": "Data Farming Active Rewards APY resets each Round and is relative to your Ocean locked and DataNFT curation.\n\nThe longer you lock, and the better you curate, the higher your APY will be.",
+  "tooltip_rewards_history_active_rewards": "Rewards distributed for active curation of datasets. Active rewards are emitted based on Data Consume Volume.<br/><br/>Please check the docs for more info.",
+  "tooltip_rewards_apy_passive_history": "Average / User\n\nHistorical Passive Rewards distributed to users holding veoCEAN.",
+  "tooltip_rewards_apy_volume_history": "Average / User\n\nHistorical Volume DF Rewards distributed to users engaged in Data Farming.",
+  "tooltip_active_rewards": "The formula for calculating the active rewards can change. You can find the formula for the current round <a href=https://docs.oceanprotocol.com/veocean-data-farming-tutorial target=_blank>here</a>"
+}
```
idiom-bytes, line 45: Our linters are conflicting.
Lots of changes happening although very little changed.

idiom-bytes:
I pushed some updates for this PR inside this branch/pr.
https://github.com/oceanprotocol/df-web/pull/585

Please merge it in.
1. Please don't change headers.
2. Please adjust total claims.

idiom-bytes:
**[Rewards from config & endpoint]**
Finally, I think we're going to have to update the rewards/breakdown by requesting from df-sql.
1. Right now predict-eth awards $2500 USD in OCEAN.
2. The amounts we announce are in OCEAN.
3. This means that the reward amount in OCEAN fluctuates and isn't fixed right now, so we can't rely on epochs.json. 
4. We need to get the final implementation for this.

src/App.svelte
KatunaNorbert: It's a very small change so I wouldn't open another PR for that. This PR should be merged only when we are ready to launch so this implies that we should check nothing it's broken.

We should change from datafarming since now all of the substreams are part of data farming. We could use another name but I'd like to be consistent, so the name we are displaying should be the name used internally so we It won't cause misunderstandings or mistakes.

src/utils/metadata/descriptions.json
KatunaNorbert: Yeah, should setup a linter config file, maybe once WC PR it's merged because that contains a new module bundler

idiom-bytes:
Removed code that filters by userAddress:
1. It was never used to pass a userAddress.
2. Wherever it was used, it was passing in an empty args.
3. This is not part of the requirements for the history table.
4. This function is now being shared between logic that should not be shared. User History should not be shared with Global History, why? Because if the user is logged in, we still need to use global reward history.

If we need it back, please implement under a separate function so user-history is not mixed-up with global history. I.E.
1. getChallengeDFUserRewards
2. getChallengeDFGlobalRewards

```
export const getChallengeRewards = async (userAddress) => {
  let res;
  try {
    res = await fetch(`${import.meta.env.VITE_BACKEND_API}/challenge/rewards`, {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      body: JSON.stringify({
        query: userAddress
          ? {
              round: {
                $gt: -1,
              },
              winner_addr: userAddress.toLowerCase(),
            }
          : {
              round: {
                $gt: -1,
              },
            },
        fields: ["winner_addr", "round", "OCEAN_amt"],
      }),
    });
  } catch (error) {
    console.log(error);
    return 0;
  }
  let data = await res.json();
  return data;
};
```

PR #583, opened 2023-06-16
Create "Challenges" page.

### LINKED ISSUE #582: Implement "Challenges" Page
### DoD: Challenge page has been built and integrated
- [x] Challenge page is up and running
- [x] "Active Challenge" card is shown and links to predict-eth repo
- [x] predict-eth repo has a `df-challenge.md` or similar file that is fixed/constant
- [x] There is a "Past Challenges" card that has a table of recent winners
- [x] Table of winners include top 3 prizes, and links to each user on etherscan
- [ ] Table retrieves results from df-sql and serves it under "Past Challenges"
- [ ] Rewards page links to "Challenge" page
- [ ] Other source-links point to the "Challenge" page (such as oceanprotocol.com)

idiom-bytes:
Fixes #582
- DoD, requirements, and comments can be found in issue.

Latest screenshot:
Screenshot from 2023-06-15 18-33-40

### DIFF
--- a/src/stores/data.js
+++ b/src/stores/data.js
@@ -8,7 +8,7 @@ import { getRoundsDatafarm } from "../utils/functions";
 let networksData = networksDataArray.default
 
 export let datasets = writable("");
-export let dataChallenges = writable("");
+export let dataChallenges = writable([]);
 
 export const columnsData = {
   'alloc': [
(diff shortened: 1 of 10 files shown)


src/stores/data.js
```diff
@@ -271,3 +272,111 @@ export async function loadDatasets(nftsApi, allocations) {
...
+    return [];
+  }
+  let data = await res.json();
+  return data;
+}
+
+const dfChallengeRewardTableMaxRounds = 10;
+const dfChallengeRewardBaseRound = 5;
+
+export async function loadDFChallengeResults() {
```
idiom-bytes, line 330: Tagging.

KatunaNorbert:
The nav bar menu it's getting to crowded now, we should do something about it

KatunaNorbert:
Did some changes:
- Made nav menu responsive since now we have more items and it breaks the interface on smaller screens
- Active challenge card was breaking the interface on smaller screens
- Removed card from table since we don't have one around any table we use, it gives more space to table and looks better
- Removed table description, it looked pretty redundant and I think the table it's self explanatory, we can add it back If really needed
- Moved the winners wallet address under the link, so user can also click the address instead of just the icon
- Added skeleton component to table so the UI is consistent and we don't have to show the loading message but rather the skeleton table with headers visible
- Added skeleton table component to data farming table

KatunaNorbert:
We should present the rewards for each place so it's clear for the users that only first 3 of them are going to receive rewards and the amount each place it's going to receive, something like this:

Screenshot 2023-06-23 at 15 29 00

Then we can remove the total rewards since that's obvious and now we have a more detailed overview about the rewards and how they are distributed.

idiom-bytes:
```Then we can remove the total rewards since that's obvious and now we have a more detailed overview about the rewards and how they are distributed.```

I disagree and don't believe we should make this change.
1. In any CTA you don't want to overwhelm the user, [keep it brain-1](https://en.wikipedia.org/wiki/Thinking,_Fast_and_Slow).
2. The table below offers all the information needed (making the visual redundant).

KatunaNorbert:
@idiom-bytes the table offers information about past challenges. If rewards amount change then what should the user assume for the number of rewards for each place?
I'd like for each challenge to be explicit and not relay on checking past challenges or other places to understand information, and for user to assume that current challenge total reward it's going to be split up for each place based on the last one.
More than that the table doesn't show the total rewards, so user has to sum them up and then assume that current challenge it's the same as last one.

KatunaNorbert (approved):

KatunaNorbert:
Another change I made is link each past challenge to the right description from predict-eth repo. For this I updated the round to contain the name and transformed it to a link. We might keep it as Round or change table name to Name so it's more linked to the current challenge name.

The active challenge should also contain the Round in the name and now can be consistent to the rest of the history table content but also with the name presented in the predict-eth repo. I added that but right now it's static, we should make it dynamic maybe based on the last round number from the history of challenges.

In the end we should have the consistent name format in the history table, active challenge and predict-eth repo, which is **Predict-ETH Round n**

idiom-bytes:
```but also with the name presented in the predict-eth repo```
https://github.com/oceanprotocol/predict-eth/pull/139

There is a PR in predict-eth where there is only going to be 1 repeating readme file. It will not change.

I will adjust the update as required.

PR #587, opened 2023-06-26
Reconcile usd ocean streams

### LINKED ISSUE #586: Reconcile USD predictETH with OCEAN streams
[PredictETH /ChallengeDF - USD vs. OCEAN]

### Problem:
Right now Predict-ETH / announcements all communicate $2500 USD Rewards.
However, our website all communicate rewards & streams in OCEAN.

We need to reconcile this in the FE which is a bit of a challenge. Example:
**[Beginning of Week]**
We need to know this info 1-week upfront (how much OCEAN is going to go in each stream)

**[End of Week]**
We only get the reward results from df-sql at the very end of the week (how much OCEAN was actually paid out).
In this case, if we wanted to be accuracte, we may report the rewards/stream with => $2500*OCEAN_USD being updated
This means that this number would keep changing

### DoD:
- [ ] Reconcile this design
- [ ] Implement the solution

idiom-bytes:
Fixes #586 
- Structures streams & epoch file such that the required parameters exist
- Recalculates rewards by taking ChallengeDF rewards from the top,
- Reports USD equivalent values wherever appropriate (ChallengeDF only right now)

Does not fix:
- Broad usage of magic numbers to index passive/active/other rewards (address in separate issue)

### DIFF
--- a/src/components/claim/ClaimPortal.svelte
+++ b/src/components/claim/ClaimPortal.svelte
@@ -21,6 +21,7 @@
   import { oceanUnlockDate } from "../../stores/veOcean";
   import { getRate } from "../../utils/market_rates";
   import * as streamsData from "../../utils/metadata/rewards/streams.json"
+  import { onMount } from "svelte";
 
   let loading = false;
   let veBalance = 0.0;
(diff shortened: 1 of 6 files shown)


KatunaNorbert:
If rewards are going to be displayed in USD but distributed in OCEAN we should also update the Challenges history table to also contain the rewards in OCEAN

KatunaNorbert:
I really don't like all the rewards being updated and fluctuate trough the entire week just because we want challenges to have a fixed USD values, but If we really want to stick with this we should explain this with a message so user has an idea about the rewards amount fluctuation.

KatunaNorbert (approved):
LGTM, just did some small cleanup

idiom-bytes:
```
<h4>{`${title} - ${availableRewards} OCEAN ${availableRewardsUSD ? `- (${availableRewardsUSD} USD)` : ''}`}</h4>
```

I know this isn't python but I created that basic if/else so w we can reduce some of these line lengths.
Sometimes, having a few more lines of code makes everything more legible and easier to update.

PR #596, opened 2023-06-28
Fetch challenges data from df-sql

### LINKED ISSUE #595: Load Challenge DF related data from df-sql
Replace following mocked data with data provided by df-sql:
- [x] display key metric inside the Challenge DF substream
- [x] display winner address for each round

KatunaNorbert:
Fixes #595.

### DIFF
--- a/src/utils/rewards.js
+++ b/src/utils/rewards.js
@@ -217,6 +217,38 @@ export const getDFallocations = async (userAddress) => {
   return data;
 };
 
+export const getChallengeRewards = async (userAddress) => {
+  let res;
+  try {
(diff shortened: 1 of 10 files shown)


src/utils/rewards.js
```diff
@@ -217,6 +217,38 @@ export const getDFallocations = async (userAddress) => {
   return data;
 };
 
+export const getChallengeRewards = async (userAddress) => {
+  let res;
+  try {
+    res = await fetch(`${import.meta.env.VITE_BACKEND_API}/challenge/rewards`, {
```
idiom-bytes, line 223: Looks great!

PR #600, opened 2023-06-29
Display substreams unclaimed rewards

### LINKED ISSUE #599: Display claimable rewards broken down by stream
In order to complete this we may need:
- [ ] To know when the user claimed last active rewards relative to distributions (passive contract)
- [ ] To know when the user claimed last passive rewards relative to distributions (active contract)
- [ ] To know how much the user earned per stream per round
- [ ] We can then break down the claimable amount based on each stream

KatunaNorbert:
Fixes #599.

### DIFF
--- a/src/components/claim/ClaimPortal.svelte
+++ b/src/components/claim/ClaimPortal.svelte
@@ -8,17 +8,20 @@
     veClaimables,
     dfClaimables,
     getDFRewards,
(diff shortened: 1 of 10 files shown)


idiom-bytes (approved):
Looks good. Pushed a minor spelling fix.

PR #604, opened 2023-06-30
combine prize and winner column

### LINKED ISSUE #601: Group prizes and winners inside challenges history

KatunaNorbert:
Fixes #601.

### DIFF
--- a/global.css
+++ b/global.css
@@ -200,16 +200,17 @@ body button:focus {
   width: 100vw !important;
 }
 
-.bx--toast-notification__caption{
+.bx--toast-notification__caption {
   display: none;
 }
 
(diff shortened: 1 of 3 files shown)


idiom-bytes (changes requested):
The only problem I see is that:
1. Ocean reward + address are combined
2. URL links to address based on the string used
3. User is then sent to a bogus address on etherscan (https://etherscan.io/address/1,500%20OCEAN%20-%200xabcd...cdef01)

How to fix: Make it so that only the user address is used in the `etherscan.io/address/ahref`

idiom-bytes (approved):
Looks good!

PR #628, opened 2023-08-11
Display challenge winners's nmse

### LINKED ISSUE #627: Display winner's NMSE
This is a request from community.
Inside the winners table display winners NMSE alongside the address so users can check how far they were.

data available here:
'https://df-sql.oceandao.org/challenge/data' you can call it by passing the round number in the request body similar to rewards query

KatunaNorbert:
Fixes #627.


idiom-bytes (approved):
Reviewed, looks good.

PR #630, opened 2023-08-17
Restructure navbar&titles

### LINKED ISSUE #629: Restructure the nav bar / titles
### Problem
There are various improvements in the DF page that we should implement.
Let's restructure the Nav Bar + Pages a bit so they are easier to follow

### DoD:
- [x] Rename veOCEAN => PASSIVE
- [x] Move `Delegate` inside of `Volume-DF` (bottom of page)
- [x] Top Nav Bar should be: [PASSIVE, VOLUME DF, CHALLENGE DF, REWARDS, ABOUT]
- [ ] 1 title + 1 sentence description of how each page works

KatunaNorbert:
Fixes #629.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -227,7 +227,10 @@
     <Route path="/passive-df" primary={false}>
       <VeOceanPortal />
     </Route>
-     <Route path="/veocean" primary={false}>
+    <Route path="/delegate" primary={false}>
+      <VeOceanPortal />
+    </Route>
+    <Route path="/veocean" primary={false}>
       <VeOceanPortal />
     </Route>
     <Route path="/challenge-df" primary={false}>
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -224,7 +224,10 @@
     <Route path="/volume-df" primary={false}>
       <DataPortal />
     </Route>
-    <Route path="/passive" primary={false}>
+    <Route path="/passive-df" primary={false}>
+      <VeOceanPortal />
+    </Route>
+     <Route path="/veocean" primary={false}>
       <VeOceanPortal />
     </Route>
     <Route path="/challenge-df" primary={false}>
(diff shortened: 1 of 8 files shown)


src/App.svelte
```diff
@@ -225,11 +224,11 @@
     <Route path="/volume-df" primary={false}>
       <DataPortal />
     </Route>
-    <Route path="/veocean" primary={false}>
+    <Route path="/passive-df" primary={false}>
       <VeOceanPortal />
     </Route>
-    <Route path="/delegate" primary={false}>
-      <DelegationPortal />
```
idiom-bytes, line 232: Hi Norbert, please do not delete routes.
Simply update the route and push the user to the right place.

idiom-bytes:
Hi Norbert, looks good. Some notes:
- Please, don't delete routes, just update them to send the user to the right place.
- Please, don't use `&` on the git commit name because it causes all sorts of issues :sweat_smile: 
- I see what you're doing with the subtitle in volume-df page, but we need to adjust it. I'm going to do a quick fix here, and then create a separate issue to adjust these things.

Screenshot from 2023-08-21 10-11-26
_Now the sentence doesn't describe the feature, and instead tells the user about the delegation feature_

idiom-bytes (approved):
Looks good.

idiom-bytes (approved):
lgtm

src/App.svelte
```diff
@@ -227,7 +227,10 @@
     <Route path="/passive-df" primary={false}>
       <VeOceanPortal />
     </Route>
-     <Route path="/veocean" primary={false}>
+    <Route path="/delegate" primary={false}>
+      <VeOceanPortal />
```
KatunaNorbert, line 231: I think this should point to `DataPortal`, volume page

idiom-bytes (approved):
lgtm

PR #632, opened 2023-08-17
Updated app banner

### LINKED ISSUE #631: Remove 2 existing banners (DF main, 2x stake), add new banner (Challenge DF)
### Background / motivation

There are two banners at the top of df.oceandao.org right now:
- "Data Farming Main has started! Read [the blogpost](https://blog.oceanprotocol.com/ocean-data-farming-main-is-here-49c99602419e) for more information."
- "Publish your datasets through our [marketplace](https://market.oceanprotocol.com/publish/1) to get [2X Stake](https://blog.oceanprotocol.com/data-farming-publisher-rewards-f2639525e508) for publishers."

(Pic below)

### Issues

Main issues:
- Current banners are old. DF Main was Mar 16, 2023 (4 mos ago), and the 2X stake thing was Feb 16, 2023 (6 mos ago)
- Those two banners distract from more useful information. "About" points to all the info that "DF main" has. "2X stake" is only for volume DF, and we have near-zero volume anyway
- The existence of two banners can make us reluctant to add more features yet, ones that might actually help UX
- _Two_ banners at once is too confusing. We really should have never done that. One is enough, eg like in oceanprotocol.com for years. So if we have banners in the future, let's restrict to one. I expect that we _will_ have banners in place
- We just released Challenge DF two weeks ago. We really should have a banner for that, for now

### TODO

- Remove both existing banners (DF main, 2X)
- Add banner for Challenge DF 
- Keep infrastructure in place to readily add one banner, but not two

### Screenshot of status quo
df

KatunaNorbert:
Fixes #631.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -209,12 +209,8 @@
 
 <Router>
   <BannerMessage
-    title={`Data Farming Main has started!</a>`}
-    message={`Read <a href='https://blog.oceanprotocol.com/ocean-data-farming-main-is-here-49c99602419e' target='_blank'><strong>the blogpost</strong></a> for more information.`}
-  />
-  <BannerMessage
-    type="default"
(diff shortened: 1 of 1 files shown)


idiom-bytes (approved):
Looks great.

PR #635, opened 2023-08-21
Add predictoor substream

### LINKED ISSUE #634: Add Predictoor substream
Df-web needs to be updated and have a new substream for Predictoor. 

All data needed for Predictoor component exists inside of df-sql. We'll display what's possible, and a brief summary.
- Do not add Oasis network/claiming from df-web
- Point the user to `pdr-backend` readme
- The following can be built using data that will exist in df-sql (predictoorData + predictoorSummary)

image

KatunaNorbert:
Fixes #634.


src/utils/predictoor.js
```diff
@@ -0,0 +1,26 @@
...
+            "accuracy",
+          ]
+        }),
+      });
+    } catch (error) {
+      console.log(error);
+      return 0;
+    }
+    let data = await res.json();
+    return data.length > 0 ? data.legth : 0
```
idiom-bytes, line 25: Pretty sure this query & results won't work.
1. You can't get accuracy from `/predictoor/data`. You should use predictoorSummary instead.
2. You're returning the length of the result, not the accuracy of the user. If you use the method above (1) you're also only going to report `per-round-accuracy` (i.e. accuracy of the last round), rather than `accuracy-average-lifetime`.

image

I would recommend we report a simple "prediction record" as the metric.

`${sum(correct_prediction_count)} / ${sum(total_predictions)}`

idiom-bytes (dismissed):
Please review comments on query & results.

idiom-bytes:
@KatunaNorbert I see you made updates to this PR but you haven't resolved my comments.
Is this ready to go?

src/utils/predictoor.js
KatunaNorbert: I get the data from `/predictoor/summary`, and it displays as you mentioned the last round accuracy.
Maybe we can discuss a little bit if the metric should be lifetime average or per round. I personally think per round is more valuable because you can then compare it to previous rounds, or with other users, as for a lifetime metric is hard to tell how you performed during a round.
I would also like to have the number of correct predictions / total predictions because it's not really valuable for me if I have 70% accuracy but only 50 predictions to compare with someone else who has 65% but 2000 predictions only based on accuracy, the number of predictions also matter.

idiom-bytes (approved):
lgtm

PR #646, opened 2023-10-12
Add support for predictoor assets

### LINKED ISSUE #645: Support Predictoor assets

KatunaNorbert:
Fixes #645.


src/stores/web3.js
```diff
@@ -58,6 +58,7 @@ web3Modal.setTheme({
 // TODO - Replace networkData w/ networksDataArray
 export function getNetworkDataById(data, networkId) {
   if (!networkId) return;
+  console.log
```
kdetry, line 61: we can remove it I guess

kdetry (changes requested):

src/stores/web3.js
KatunaNorbert: yes

kdetry (approved):

PR #647, opened 2023-10-16
Handle multiple claims for Passive Rewards

### LINKED ISSUE #637: Fix Passive Rewards claimable amount
FeeEstimate contract passive rewards calculations only returns the rewards for the first claim. If user needs to claim multiple times the rewards for the other claims are not visible.
Calculate unclaimed rewards on the FE using df-sql and subgraph.

KatunaNorbert:
Fixes #637.

### DIFF
--- a/src/components/claim/ClaimRewardsVeDF.svelte
+++ b/src/components/claim/ClaimRewardsVeDF.svelte
@@ -11,6 +11,7 @@
     APYs,
     lastActiveRewardsClaimRound,
     oceanUserRewards,
+    lastPassiveRewardsClaimRound,
   } from "../../stores/airdrops";
   import ClaimItem from "../common/ClaimItem.svelte";
   import Swal from "sweetalert2";
@@ -26,6 +27,7 @@
   export let canClaimDF = true;
   export let streams;
   let claiming;
+  let passiveRewards = 0;
   const supportedChainId = import.meta.env.VITE_VE_SUPPORTED_CHAINID
 
   async function onClaimDfRewards() {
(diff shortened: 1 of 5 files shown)


src/components/claim/ClaimRewardsVeDF.svelte
```diff
@@ -146,11 +148,24 @@
...
+      $oceanUserRewards.forEach((r) => {
+        if(r.round >= $lastPassiveRewardsClaimRound) {
+          sum += r['sum(passive_amt)']
+        }
+      })
+      passiveRewards = sum
+      streams[0].substreams[0].availableRewards = sum
+  }
+
+  $:if($lastActiveRewardsClaimRound >= 0 && $oceanUserRewards) calculateUnclaimedPassiveReward()
```
idiom-bytes, line 162: Hi, just to be clear, it seems like the logic is now aggregating all claimable rewards.

This is different than what I heard in the stand up, where only the next one (rather than multiples) will be visible to the user.

If all rewards are displayed and user can start claiming them, then great!

idiom-bytes: I left some comments here:
https://github.com/oceanprotocol/df-web/issues/637#issuecomment-1765325539

I think what you mean is "Next Claim {amount}" rather than "First Claim {amount}".

If correct, please adjust the button text to be more intuitive.

KatunaNorbert: I updated based on the discussion from within the issue

idiom-bytes (approved):
Looks great!

idiom-bytes (approved):
LGTM

PR #661, opened 2023-11-02
Allocations action buttons updates and fixes

### LINKED ISSUE #455: In Rewards page: disable "update allocations" button if no changes yet
When I go to the "rewards" page, I see the "Update allocations" active and if I press it without changing anything it will show a modal with an error message. Can we instead disable this button if there's no change yet?

Before pressed:
Screenshot 2023-01-30 at 11 59 44

After pressed:
Screenshot 2023-01-30 at 11 59 35

ALSO the "reset" button looks disabled, but you can actually press on it. Can we give it another style? Right now it's kind of confusing to me. 

Screenshot 2023-02-01 at 09 08 03

KatunaNorbert:
Fixes #455.


idiom-bytes (approved):
Read comments, simple solution!

PR #668, opened 2023-11-07
Use only 1 chain for Wagmi public client

### LINKED ISSUE #667: Wallet connect user only 1 chain for public client

KatunaNorbert:
Fixes #667.


KatunaNorbert:
@idiom-bytes this fixes the different deployment issues because of contract call if wallet not connected and now APY should be displayed every time.

kdetry (approved):

kdetry:
I approve if the package-lock change is okay

PR #672, opened 2023-11-10
Fix Coinbase wallet connection

### LINKED ISSUE #666: Users are reporting issue connecting to dApp w/ Coinbase wallet
### Problem
As reported on discord, an user seems unable to connect to dapp w/ CB wallet.
https://discord.com/channels/612953348487905282/1169308989113249813/1170343667949650040

### DoD
- [ ] Users are able to connect and interact w/ df dapp via CB wallet

KatunaNorbert:
Fixes #666.


kdetry:
It works so I approve it. But there is no logo at the coinbase option. I think it is not important but if we can, it would be better.

kdetry (approved):

KatunaNorbert:
Looks to be some kind of issues with the logo image link inside the package. I thought it's better to relay on the image provided by the package although it's not working now because If we use an external link it might not be reliable enough

PR #673, opened 2023-11-14
Add apy calculator to lock component

### LINKED ISSUE #612: [APY calc v1] Provide a simple "Estimate APY vs. Best-case APY" calculation in the OceanLockComponent.
### Problem:
- We do not want to implement a full calculator in the FE
- We do want to show the user how much better they could do if they locked for 4 years
- We want to provide users with enough context so they feel like they are intentionally making this decision ("Try to remove foot-gun conditions")
- Provide a tooltip with clear reference to FinePrint(tm)

image (3)

### FinePrint(tm)
1. Assuming a 4-year lock up period.
2. Assuming a weekly schedule of compounding rewards into an updated lock.
3. Assuming current: number of users, reward emissions, and other key metrics remain constant.
4. Excluding lock, claim, and any tx fees.

Final Text (and calculation until further adjustments):
```
APYs are calculated assuming an initial 4-year lock up period with a weekly schedule of compounding rewards into an updated 4-year lock. This estimate works provided current: number of users, reward emissions, and other reward parameters stay constant while excluding all tx fees.
```

### DoD:
- [ ] User chooses his lockup period via shortcuts or calendar
- [ ] User sees his estimated APY based on his decisions
- [ ] User sees the estimated Best-Case APY (4yr, compound, etc...)
- [ ] User is provided with a message that tells him "Your APY could be X amount larger if you locked for 4 years"
- [ ] Use the same information as the "Average APY calculations" to estimate this number (i.e. currentEpoch, or lastEpoch)

KatunaNorbert:
Fixes #612 .


kdetry:
LGTM

kdetry (dismissed):

kdetry (approved):

PR #675, opened 2023-11-20
Fix Predictoor accuracy metric

### LINKED ISSUE #674: Predictoor-DF substream last round accuracy metric improvements
- [ ] format the accuracy to only have a fixed decimal format
- [ ] switch to next round accuracy only after ROSE rewards are distributed(that is when the data for the previous round is available)

KatunaNorbert:
Fixes #674.


idiom-bytes (approved):
Float operation looks good.

PR #685, opened 2023-11-29
Add compounding switch

### LINKED ISSUE #681: Add component for turning on/off optimal compounding
Inside the Lock Ocean component, in the area where the APY and rewards are displayed, add a component that can switch the optimal lock configuration function on and off. 
The component should be suggestive and contain the information necessary for the user to understand what is the purpose of it.
The default state is going to be 'on'

KatunaNorbert:
Fixes #681.


src/components/common/Switch.svelte
```diff
@@ -0,0 +1,239 @@
...
+	export let value = 'on';
+
+    let checked = true;
+	const uniqueID = Math.floor(Math.random() * 100)
+
+    function handleClick(event){
+        const target = event.target
+        const state = target.getAttribute('aria-checked')
+        checked = state === 'true' ? false : true
+        value = checked === true ? 'on' : 'off'
```
kdetry, line 15: if the aria-checked value is 'true', the checked variable becomes false. It made me confused.  
and this line can be `checked ? 'on' : 'off' `

I do not have anything else about the code, thank you

PR #686, opened 2023-11-29
issue682 optimal compound part 1

### LINKED ISSUE #682: Implement function for calculating most optimal lock configuration
The function should use the lock ocean form input values(amount, date) and figure out the most optimal amount of compounds for the best yield.

kdetry:
Fixes #682 
It is just a draft for now

### DIFF
--- a/src/components/veocean/LockOcean.svelte
+++ b/src/components/veocean/LockOcean.svelte
@@ -3,7 +3,7 @@
     userAddress,
     connectedChainId,
     switchWalletNetwork,
-    getNetworkDataById
+    getNetworkDataById,
   } from "../../stores/web3";
   import Button from "../common/Button.svelte";
   import Card from "../common/Card.svelte";
@@ -12,12 +12,9 @@
   import ItemWithLabel from "../common/ItemWithLabel.svelte";
   import TokenApproval from "../common/TokenApproval.svelte";
   import AgreementCheckbox from "../common/AgreementCheckbox.svelte";
+  import { allowance, getTokenPriceFromCoingecko } from "../../utils/tokens";
   import {
-    allowance, getTokenPriceFromCoingecko
-
-  } from "../../utils/tokens";
-  import {
-  ethPrice,
+    ethPrice,
     getOceanBalance,
     oceanPrice,
     updateUserBalanceOcean,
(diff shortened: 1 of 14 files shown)


src/utils/rewards.js
```diff
@@ -75,19 +77,174 @@ export const getPassiveUserAPY = async (userVeOcean, lockedOcean) => {
...
+  }
+
+  let principal = startPrincipal;
+
+  let totalRewards = 0;
+
+  let totalVotingPower = 0;
+  for (let i = 0; i < compounds; i++) {
+    let tempVotingPower = getVotingPower(
+      msDelta / compounds,
```
KatunaNorbert, line 157: we shouldn't divide the lock time by compounds because it's going to result wrong veOCEAN amount

src/components/veocean/LockOcean.svelte
```diff
@@ -241,56 +244,143 @@
...
+      lockedOceanAmount: $lockedOceanAmount,
+      formAmount: $form.amount,
+      formUnlockDate: $form.unlockDate,
+      fees,
+      totalVeOceanSupply: $totalVeOceanSupply,
+    });
+
+    console.log('optimumComp',optimumComp)
+
+    for (let i = 1; i <= MAX_COMPOUNDS; i++) {
```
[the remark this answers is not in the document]
kdetry: yes, it was just for test, deleted

src/utils/rewards.js
kdetry: Firstly, It is moved to this part:
https://github.com/oceanprotocol/df-web/pull/686/files#diff-b357279e0988e4f14f3ba52534dbc149574f1aecaab34c70dcbe95276e863124R30

However, this modification doesn't negatively impact the calculation. To imagine, let's consider a scenario where maxDate equals 'X'.

When we divide 'X' by 3, for instance, we essentially have three separate segments, each representing `(1/X) * totalAmount`. this is expressed as 3 * (1/X) * r. On the other hand, if we didn't divide and directly used `(3/X) * r`, the outcome remains the same.

The division of the lock time by the number of compounds is crucial because the `totalAmount` changes with each compounding event. Therefore, to accurately account for these changes over time, the lock time must be divided accordingly

kdetry:
I deleted every part which affects the UI, it is currently only the calculation step of the feature

src/utils/compound.js
```diff
@@ -0,0 +1,149 @@
...
+
+      const rawPeriodReward = periodReward;
+
+      if (i === 0) {
+        periodReward -= fees.lock;
+      }
+
+      if (i === currentCompoundCount - 1) {
+        periodReward -= totalPeriodClaims;
+        periodReward -= fees.withdraw;
```
KatunaNorbert, line 59: I don't think we need to subtract the lock and withdraw fees from specific periods. Those should not impact the period rewards but only the entire lock rewards.

kdetry:
both of us have worked on the same PR
https://github.com/oceanprotocol/df-web/pull/689

PR #659, opened 2023-11-01
Rename Volume DF table header terms

KatunaNorbert:
Fixes #655.

PR #696, opened 2023-12-13
removed challenge df related logic

KatunaNorbert:
Fixes #659.


idiom-bytes:
I reviewed the PR and it looks good.

I also tried to further remove stuff from terms.md, EpochHistory.svelte, and rewards.js, however I decided to leave those there.

I do think that ChallengeDF has sunset and we want users to focus on what's next (becoming a predictoor) and that is now somewhat missing.
- The banner are for announcements/new things
- The rewards page, only have 1 line and 1 button to take you to predictoor
image

Therefore, although this PR looks good, I think we lost a good amount of visibility/prevalence towards the reward program. I.E. How can we perhaps make predictoor rewards more prevalent?

idiom-bytes (approved):
Looks good to merge.

Left feedback for how to keep improving this.

